PR #2, opened 2022-03-03
DAG-1668: Generate fuzzer task configurations

dbradf:


src/evg_config.rs
```diff
@@ -0,0 +1,848 @@
...
+        self.evg_project.task_def_map()
+    }
+
+    /// Get a list of build variants with the required build variants at the start.
+    fn sort_build_variants_by_required(&self) -> Vec<String> {
+        let build_variant_map = self.get_build_variant_map();
+        let mut build_variants: Vec<String> = build_variant_map
+            .keys()
+            .into_iter()
+            .filter_map(|bv| {
```
dbradf, line 69: An `Option` is an enum type in rust that is used to replace `null`. The enum looks like this (note: it uses generics to work with any types, that is the `T`):

```rust
enum Option<T> {
  Some(T),
  None,
}
```

This enforces null safety with the compiler. It is similar to the `Optional` type in python, but better enforced. 

The `filter_map` function will iterator over values, apply a function to each value and return the result as long as the result is not `None`. What we are doing here is iterating over all the build variant names and returning `Some` for any that end in "-required" and `None` for the rest. As a result, the `filter_map`, just returns us a list of required build variant.

src/evg_config.rs
```diff
@@ -0,0 +1,848 @@
...
+                if bv.ends_with(REQUIRED_PREFIX) {
+                    Some(bv.to_string())
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        build_variants.extend::<Vec<String>>(
```
dbradf, line 78: The compiler can't tell we are trying to extend with a `Vec<String>` (due to using `collect` which support many types), so we have to tell it here.

src/evg_config.rs
```diff
@@ -0,0 +1,848 @@
...
+
+    if let Some(Function(func)) = command {
+        Some(func)
+    } else {
+        None
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
dbradf, line 484: This create a new module. Since it is tagged with `#[cfg(test)]`, rust knows that it is just test code and will strip it out of the binary.

src/evg_config.rs
```diff
@@ -0,0 +1,848 @@
...
+
+    use super::*;
+
+    // is_task_generated tests.
+
+    #[test]
+    fn test_is_task_generated_should_return_false_if_not_generated() {
+        let evg_task = EvgTask {
+            commands: vec![fn_call("hello world"), fn_call("run tests")],
+            ..Default::default()
```
dbradf, line 497: rust doesn't have default parameters, but uses this `Default::default()` instead. This is a trait you can define for types that provides some default values.

src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+mod evergreen_names;
+mod evg_config;
+mod resmoke_proxy;
+mod task_name;
+mod task_types;
+
+/// Directory to store the generated configuration in.
+const CONFIG_DIR: &str = "generated_resmoke_config";
+
+lazy_static! {
```
dbradf, line 41: `lazy_static` is a macro that will let you define static values, but they won't actually be created until they first get used.

src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+    static ref EXPANSION_RE: Regex =
+        Regex::new(r"\$\{(?P<id>[a-zA-Z0-9_]+)(\|(?P<default>.*))?}").unwrap();
+}
+
+type GenTaskCollection = HashMap<String, Box<dyn GeneratedSuite>>;
```
dbradf, line 46: `type` lets you define a name more a more complex type.

src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `evg_project_location` - Path to the evergreen project configuration yaml.
+    ///
+    /// # Returns
+    ///
+    /// A set of dependencies to run against.
+    pub fn new(evg_project_location: &Path) -> Result<Self> {
+        let discovery_service = Arc::new(ResmokeProxy::new());
```
dbradf, line 65: `Arc` has to do with memory management. It stands for "Atomic reference counter". It will let multiple threads access the data in the variable. We don't actually need it right now, but we will in a future ticket.

src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+}
+
+/// Create 'generate.tasks' configuration for all generated tasks in the provided evergreen
+/// project configuration.
+///
+/// # Arguments
+///
+/// * `deps` - Dependencies needed to perform generation.
+/// * `config_location` - Pointer to S3 location that configuration will be uploaded to.
+pub fn generate_configuration(deps: Dependencies, config_location: &str) -> Result<()> {
```
dbradf, line 106: `()` is a unit type. Here it means this function doesn't return anything. In this case, the function won't return anything, but it could fail. So we want to return a `Result` in case there is an error. The `()` is used for that. If there is an error the `Result` will be an error, if everything was successful, the result will be `()`.

src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+        buildvariants: generated_build_variants.to_vec(),
+        tasks: task_defs,
+        ..Default::default()
+    };
+
+    std::fs::create_dir_all(CONFIG_DIR).unwrap();
+    let mut config_file = Path::new(CONFIG_DIR).to_path_buf();
+    config_file.push("evergreen_config.json".to_string());
+    std::fs::write(config_file, serde_json::to_string_pretty(&gen_evg_project)?)?;
+    println!("{}", serde_yaml::to_string(&gen_evg_project)?);
```
dbradf, line 134: For now, I'm both writing the generated configuration to a file and printing it out to the stdout. This is mostly for debugging purpose while the tool is being built out.

src/task_types/fuzzer_tasks.rs
```diff
@@ -0,0 +1,515 @@
...
+                        .collect::<Vec<EvgTask>>(),
+                );
+            }
+        } else {
+            sub_tasks = (0..params.num_tasks)
+                .map(|i| build_fuzzer_sub_task(&params.task_name, i, params, None, None))
+                .collect();
+        }
+
+        Ok(Box::new(FuzzerTask {
```
dbradf, line 212: A `Box` is a pointer to data. There are technical reasons why we are using it here. The summary is that we are going to treat it as a trait, but since traits are polymorphic, they need to be wrapped in some container.

src/evg_config.rs
```diff
@@ -0,0 +1,848 @@
...
+                    }
+                })
+                .collect(),
+        );
+
+        build_variants
+    }
+}
+
+/// Evaluate the evergreen and load it into a shrub project.
```
zamj, line 96: [nit] the evergreen configuration

src/evg_config.rs
```diff
@@ -0,0 +1,848 @@
...
+
+impl EvgConfigUtilsImpl {
+    /// Create a new instance of the EvgConfigUtilsImpl.
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl EvgConfigUtils for EvgConfigUtilsImpl {
+    /// Determine if the given evergreen task a generated task.
```
zamj, line 249: [nit] is a generated task.

src/evg_config.rs
zamj: Does rust usually advocate for putting the tests for a module in the same file as the code it is testing? It seems to lead to pretty large file sizes

src/lib.rs
```diff
@@ -0,0 +1,371 @@
...
+        let mut seen_tasks = HashSet::new();
+        for build_variant in &build_variant_list {
+            let build_variant = build_variant_map.get(build_variant).unwrap();
+            for task in &build_variant.tasks {
+                if !seen_tasks.contains(&task.name) {
+                    seen_tasks.insert(task.name.to_string());
+                    if let Some(task_def) = task_map.get(&task.name) {
+                        if self.evg_config_utils.is_task_generated(task_def) {
+                            if self.evg_config_utils.is_task_fuzzer(task_def) {
+                                event!(Level::INFO, "Generating fuzzer: {}", &task.name,);
```
zamj, line 241: This is pretty deep nesting and makes it a bit hard to understand whats happening here. Is there anything we can do to reduce how nested the interesting code portions are?

src/task_types/fuzzer_tasks.rs
```diff
@@ -0,0 +1,515 @@
...
+    ///
+    /// Map of arguments to pass to 'run tests' function.
+    fn build_run_tests_vars(
+        &self,
+        generated_suite_name: Option<&str>,
+        version_combination: Option<&str>,
+    ) -> HashMap<String, ParamValue> {
+        let mut vars = hashmap! {
+            "continue_on_failure".to_string() => ParamValue::from(self.continue_on_failure),
+            "gen_task_config_location".to_string() => ParamValue::from(self.config_location.as_str()),
```
zamj, line 86: Is this how enums are used in rust? If these are just strings, would you mind making them enums or constants?

src/evg_config.rs
dbradf: Yes, I'm following the conventions from [here](https://doc.rust-lang.org/book/ch11-03-test-organization.html). This file in particular is doing a lot, I think we can split it up.

src/evg_config.rs
dbradf: done.

src/evg_config.rs
dbradf: done.

src/task_types/fuzzer_tasks.rs
dbradf: I switched these to constants. I'm not sure I understand your question about enums.

src/lib.rs
dbradf: I broke up this a little and did some refactoring to reduce the nesting some. Hopefully that makes it a little easier to follow.

One think to note, the next ticket (adding support for generating resmoke tasks) will have a pretty big impact on this section of the code. So there are likely more changes coming in the future.

src/task_types/fuzzer_tasks.rs
zamj: Sounds good, we clarified on enums over Slack

zamj (approved):

dbradf:
evergreen merge

PR #3, opened 2022-03-09
DAG-1669: Add support for generating resmoke tasks

dbradf:


src/evergreen/evg_config.rs
```diff
@@ -5,7 +5,7 @@ use shrub_rs::models::{project::EvgProject, task::EvgTask, variant::BuildVariant
 
 const REQUIRED_PREFIX: &str = "-required";
 
-pub trait EvgConfigService {
+pub trait EvgConfigService: Sync + Send {
```
dbradf, line 8: `Send` and `Sync` have to do with multi-threading, which we are not using yet, but we will be soon. There are details in [the rust book](https://doc.rust-lang.org/book/ch16-04-extensible-concurrency-sync-and-send.html).

src/evergreen/evg_task_history.rs
```diff
@@ -0,0 +1,331 @@
...
+pub struct HookRuntimeHistory {
+    /// Name of test that hook ran with.
+    pub test_name: String,
+    /// Name of hook.
+    pub hook_name: String,
+    /// Average runtime of hook.
+    pub average_runtime: f64,
+}
+
+impl Display for HookRuntimeHistory {
```
dbradf, line 25: The `Display` trait is similar to overriding `__str__` in a python class. It allows you to customize a printable version of the object.

dbradf:
Note: There is still more work to do on the ticket that I will at in future PRs, but I thought this would be a good point for a code review.

Notably what is missing:
* Generating resmoke configuration files.
* Multi-threading task creation.
* Creating '_misc' sub-suites.

dbradf:
evergreen merge

PR #4, opened 2022-03-10
DAG-1669: Create resmoke suite files for generated sub-tasks

dbradf:


src/resmoke/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod resmoke_proxy;
```
dbradf, line 1: I pulled all the resmoke code into a new module to split it up some.

src/resmoke/resmoke_proxy.rs
```diff
@@ -0,0 +1,127 @@
+use std::{path::Path, str::FromStr, time::Instant};
```
dbradf, line 1: This code hasn't changed, it was just moved and then split out from the other resmoke suite configuration code.

src/resmoke/resmoke_suite.rs
```diff
@@ -0,0 +1,369 @@
+//! Representation of a resmoke suite file.
```
dbradf, line 1: This code was also moved and split out from the old resmoke code, but it was changed a lot. I switched from using a generic yaml loader for the resmoke configs to using serde. It cleaned up the code a lot. Instead of having super deeply nested queries about the yaml structure, we just define `struct`s the model the structure and work with them directly.

zamj: Nice, this looks a lot cleaner

zamj (approved):

dbradf:
evergreen merge

PR #10, opened 2022-03-16
DAG-1670: Add support for generating multiversion tasks

dbradf:


src/task_types/resmoke_tasks.rs
```diff
@@ -320,13 +339,61 @@ impl GenResmokeTaskServiceImpl {
...
+    /// * `sub_suites` - Sub-suites that were created for the task being generated.
+    ///
+    /// # Returns
+    ///
+    /// List of sub-suites that includes versions fall all multiversion combinations.
+    fn create_multiversion_combinations(
+        &self,
+        params: &ResmokeGenParams,
+        sub_suites: &[SubSuite],
+    ) -> Result<Vec<SubSuite>> {
```
dbradf, line 365: Here is where we are creating the configuration to actually generate the multiversion tasks. We will take the split up sub-tasks and create multiversion sub-tasks for each of them based on the multiversion configuration.

dbradf:
evergreen merge

PR #15, opened 2022-03-21
DAG-1711: Support repeating resmoke suites

dbradf:


src/task_types/resmoke_tasks.rs
```diff
@@ -112,9 +114,16 @@ impl ResmokeGenParams {
         } else {
             "".to_string()
         };
+
+        let repeat_arg = if let Some(repeat) = self.repeat_suites {
+            format!("--repeatSuites={}", repeat)
```
vincivans, line 119: What is the 'Some' stands for and what's the difference between this and [here](https://github.com/mongodb/mongo-task-generator/blob/cd62ba926bde4bb3de29b0301c87fec8f324603b/src/task_types/resmoke_tasks.rs#L93)? It seems the latter use a "&", but string does not require it?

vincivans (approved):
LGTM. I don't have objections with the logic. Maybe because I still haven't finish reading all Rust syntax, I have a general question with the syntax. I made a comment above.

src/task_types/resmoke_tasks.rs
dbradf: `Some` is part of rust's `Option` value. In this context, `repeat_suites` may or may not contain a value. If it does contain a value, that value is wrapped in a `Some`. If it does not contain a value it is `None`. The `if let` syntax basically says, if `repeat_suites` does contain a value (that is wrapped in `Some`), then the `if` statement goes down the `true` path with the value of `repeat_suites` in the `repeat` variable. If the value of `repeat_suites` is `None`, then the false path of the `if` is taken.

See the following for more details:
* [Rust Enums](https://doc.rust-lang.org/book/ch06-01-defining-an-enum.html)
* [Option Enum type: Some and None](https://doc.rust-lang.org/book/ch06-01-defining-an-enum.html#the-option-enum-and-its-advantages-over-null-values)
* [if let syntax](https://doc.rust-lang.org/book/ch06-03-if-let.html)


The `&` has to do with has rust manages memory. All variables in rust must have 1 and only 1 owner. If any other pieces of code want to use a variable, it must "borrow" the value. This is indicated with the `&`. In this case, `self` is already borrowed in this function (see the functions parameter list), so it is does not need to be borrowed again. Looking at the linked code, it appears to be in the same situations. It doesn't look like it needs to borrow the value, but doing so doesn't hurt either.

PR #16, opened 2022-03-22
DAG-1693: Add integration tests

dbradf:
This adds 2 new tests alongs with some infrastructure to mock some things out.

* The first test is an integration test that runs againsta snapshot of the server codebase's evergreen.yml and tries to generate all the configuration for it. 
* The second test uses a tool to make sure that the Cargo.toml version and the changelog have been updated (similar to what we do in other tools).

### DIFF
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -1,0 +1,30 @@
+use assert_cmd::Command;
+use tempdir::TempDir;
+
+#[test]
+fn test_end2end_execution() {
+    let mut cmd = Command::cargo_bin("mongo-task-generator").unwrap();
+    let tmp_dir = TempDir::new("generated_resmoke_config").unwrap();
+
+    cmd.args(&[
+        "--target-directory",
+        tmp_dir.path().to_str().unwrap(),
+        "--expansion-file",
+        "tests/data/sample_expansions.yml",
+        "--evg-project-file",
+        "tests/data/evergreen.yml",
+        "--evg-auth-file",
+        "tests/data/sample_evergreen_auth.yml",
+        "--resmoke-command",
+        "python3 tests/mocks/resmoke.py",
+        "--use-task-split-fallback",
+    ])
+    .assert()
+    .success();
+
+    let tmp_dir_path = tmp_dir.path();
+    assert!(tmp_dir_path.exists());
+
+    let files = std::fs::read_dir(tmp_dir_path).unwrap();
+    assert_eq!(1111, files.into_iter().collect::<Vec<_>>().len());
+}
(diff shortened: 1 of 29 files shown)


tests/integration_test.rs
```diff
@@ -0,0 +1,30 @@
...
+        "--use-task-split-fallback",
+    ])
+    .assert()
+    .success();
+
+    let tmp_dir_path = tmp_dir.path();
+    assert!(tmp_dir_path.exists());
+
+    let files = std::fs::read_dir(tmp_dir_path).unwrap();
+    assert_eq!(1111, files.into_iter().collect::<Vec<_>>().len());
```
vincivans, line 29: I'm not quite understand this line. Is this assert trying to verify total amount of files are not 1111? If so what is 1111 represent. If not, what is it for?

tests/mocks/resmoke.py
```diff
@@ -0,0 +1,59 @@
+"""Mock of resmoke.py for testing task generation."""
```
vincivans, line 1: This file is a mocked return value for resmoke.py? Like in our other project, something like `mock_git.return_value=abc`?

vincivans (commented):
Overall looks great. I'm looking for tags like we used in build-baron-tools to enable/disable some integration test, but it seems there is no tags here. Do we want to run integration tests every time we commit changes? Or maybe because its a different management system, unlike poetry, it would not generate such hook and run all tests before commit?

tests/mocks/resmoke.py
dbradf: This file is just a fake version of resmoke.py. You can call it with various options and it will return fake data. Since the tool gathers various information from resmoke, we need some way to return fake data for the integration tests. This file lets us do that. We point the tool at this file instead of a real instance of resmoke to gather the data it needs.

tests/integration_test.rs
vincivans: I'm wondering do we know there are exactly 1111 files being generated? Base on the number of test files under /tests/data/tests?

vincivans (approved):

dbradf:
evergreen merge

tests/integration_test.rs
dbradf: I ran the generator against the sample data to tell. 

We could calculate it, but the calculation gets complicated especially since we are running against a real evergreen.yml. But basically each remoke task will generate 6 files unless they are marked as multiversion, in which case they will generate 6 * (the number of multiversion fixtures configured). The multiversion value can be different depending on the task type. Also, there will be 1 file for the overall configuration.

dbradf:
evergreen merge

PR #18, opened 2022-03-24
DAG-1673: Add support for mongosync task generation

dbradf:


src/evergreen/evg_config_utils.rs
```diff
@@ -203,14 +203,18 @@ impl EvgConfigUtils for EvgConfigUtilsImpl {
     ///
     /// `true` if the given task is generated.
     fn is_task_generated(&self, task: &EvgTask) -> bool {
-        task.commands.iter().any(|c| {
-            if let Function(func) = c {
-                if func.func == GENERATE_RESMOKE_TASKS {
-                    return true;
+        if let Some(commands) = &task.commands {
```
dbradf, line 206: Apparently evergreen support tasks with no command list. We need handle if that happens, so a lot of these changes are for that situation.

src/resmoke/resmoke_proxy.rs
```diff
@@ -41,7 +41,7 @@ pub struct ResmokeProxy {
     /// Primary command to invoke resmoke (usually `python`).
     resmoke_cmd: String,
     /// Script to invoke resmoke.
-    resmoke_script: String,
+    resmoke_script: Vec<String>,
```
dbradf, line 44: If this script is composed of multiple arguments, then it needs to be interpreted as a vector not a string or the shell will not use it correctly.

vincivans (approved):

PR #19, opened 2022-04-04
DAG-1746: Support custom dependencies for generated tasks

dbradf:


src/configuration.rs
```diff
@@ -0,0 +1,13 @@
+// Configuration for generating evergreen tasks.
+
+#[derive(Debug, Clone)]
+pub struct GenerateConfig {
```
dbradf, line 4: I pulled the global settings into this struct instead of passing them around individually.

src/main.rs
```diff
@@ -116,3 +122,18 @@ async fn main() {
...
+/// Expand ~ and any environment variables in the given path.
+///
+/// # Arguments
+///
+/// * `path` - Path to expand.
+///
+/// # Returns
+///
+/// Path with ~ and environment variables expanded.
+fn expand_path(path: &Path) -> PathBuf {
```
dbradf, line 135: I noticed that we weren't translating `~/.evergreen.yml`, so I went ahead and added some code to fix that.

PR #23, opened 2022-04-18
DAG-1781: Enforce large_distro are defined

dbradf:


src/lib.rs
```diff
@@ -346,6 +402,54 @@ impl GenerateTasksServiceImpl {
...
+            }
+
+            if let Some(gen_task_config) = &self.gen_sub_tasks_config {
+                if gen_task_config.ignore_missing_large_distro(build_variant_name) {
+                    return Ok(None);
+                }
+            }
+
+            bail!(
+                r#"
```
zamj, line 434: Can you move this to be in a constant somewhere?

zamj (approved):
LGTM mod one comment

src/lib.rs
dbradf: No. The problem is that this is not a static string. It is a formatted string. You can see the formatting with `{build_variant_name}`. Rust is able to extract values to use from environment, but the cost is that the string needs to be inlined. In order to be able to pull something like this into a constant, we would need to pull in a 3rd party templating library, which seems expensive for what it would get us.

zamj: Ah, sorry, I missed the `{build_variant_name}` portion of this. Yes, pulling in a different library seems excessive, this is fine

dbradf:
evergreen merge

PR #25, opened 2022-04-20
DAG-1787: Filter enterprise tests for non-enterprise build variants

dbradf:


src/task_types/resmoke_tasks.rs
```diff
@@ -375,6 +395,33 @@ impl GenResmokeTaskServiceImpl {
...
+            .test_discovery
+            .discover_tests(&params.suite_name)?
+            .into_iter()
+            .filter(|s| self.fs_service.file_exists(s))
+            .collect();
+
+        if !params.is_enterprise {
+            test_list = test_list
+                .into_iter()
+                .filter(|s| !s.starts_with(&self.config.enterprise_prefix))
```
MikhailShchatko, line 418: Enterprise module prefix `src/mongo/db/modules` seems too broad and not enough specific to enterprise module itself.
https://github.com/mongodb/mongo/blob/7371fa1eb1008f2be98a1acdaa818c4ac82e9fa8/etc/evergreen_yml_components/definitions.yml#L73
```
modules:
- name: enterprise
  repo: git@github.com:10gen/mongo-enterprise-modules.git
  prefix: src/mongo/db/modules
  branch: master
```
I think it should be appended with `enterprise`, so that we filter out tests that are from `src/mongo/db/modules/enterprise` directory.

dbradf: updated.

src/evergreen/evg_config.rs
```diff
@@ -16,7 +16,7 @@ pub trait EvgConfigService: Sync + Send {
     fn sort_build_variants_by_required(&self) -> Vec<String>;
 
     /// Get the prefix of the given module.
```
MikhailShchatko, line 18: [nit] `prefix` -> `directory`

MikhailShchatko (approved):
LGTM

src/evergreen/evg_config.rs
dbradf: done.

PR #26, opened 2022-04-21
DAG-1790: Remember fixture settings and normalize test_files

dbradf:


src/task_types/resmoke_tasks.rs
```diff
@@ -505,18 +505,6 @@ impl GenResmokeTaskServiceImpl {
                     is_enterprise: params.is_enterprise,
                 });
             }
-            // Add a `_misc` sub-task to the list of tasks.
```
dbradf, line 508: We aren't currently running these in the build-variant level generation. And adding them is causing some problems that need to be investigated. I'm removing their inclusion for now (which is the same as our current behavior) and opened https://jira.mongodb.org/browse/DAG-1794 to investigate further.

src/evergreen/evg_task_history.rs
```diff
@@ -286,6 +288,21 @@ fn hook_hook_name(identifier: &str) -> &str {
...
+///
+/// # Arguments
+///
+/// * `test_file` - test file to normalize.
+///
+/// # Returns
+///
+/// Normalized test file.
+fn normalize_test_file(test_file: &str) -> String {
+    test_file.replace('\\', "/")
```
zamj, line 303: This is fine, but just to ask - is there anything built into Rust to convert paths between OS styles automatically like the Python Path library?

zamj (approved):

src/evergreen/evg_task_history.rs
dbradf: Yes, rust has [Path](https://doc.rust-lang.org/stable/std/path/struct.Path.html) and `PathBuf`s for dealing with paths. We might be able to use them here, but we need strings because we are using them as identifiers to track runtimes, not as actual paths. So converting them back and forth would be annoying. I know we do similar things with python code in some places.

dbradf:
evergreen merge

PR #27, opened 2022-04-22
DAG-1796: Use matrix suites for multiversion task generation

dbradf:


src/task_types/resmoke_config_writer.rs
```diff
@@ -39,8 +36,6 @@ struct WriteConfigActorImpl {
     /// Test discovery service.
     test_discovery: Arc<dyn TestDiscovery>,
 
-    multiversion_service: Arc<dyn MultiversionService>,
```
dbradf, line 42: This code no longer cares about multiversion. Any multiversion information it needs will be included in the data passed into it.

src/task_types/resmoke_config_writer.rs
```diff
@@ -192,27 +133,22 @@ impl WriteConfigActorImpl {
...
     fn write_sub_suites(
         &self,
         sub_suites: &[SubSuite],
-        origin_config: &ResmokeSuiteConfig,
-        target_name: &str,
+        resmoke_config_cache: &mut ResmokeConfigCache,
     ) -> Result<()> {
         let results: Result<Vec<()>> = sub_suites
             .iter()
+            .filter(|s| s.exclude_test_list.is_none())
```
dbradf, line 144: This `exclude_test_list` field is new. Instead of this code trying to create the tests to exclude for _misc suites (which gets complicated for multiversion), the code creating the splits will do it and include it in the `SubSuite` structs. This has the added benefit of letting use filter the sub-suites based on that field.

src/task_types/resmoke_config_writer.rs
```diff
@@ -351,16 +286,52 @@ impl ResmokeConfigActor for ResmokeConfigActorService {
     }
 }
 
+/// A cache for querying resmoke suite configurations.
```
dbradf, line 289: Calling out to resmoke tends to be really slow, so we want to minimize how frequently we do it. So this is just a simple cache so we only need to make the call out once (instead of for each sub-suite).

src/task_types/resmoke_tasks.rs
```diff
@@ -165,6 +165,12 @@ pub struct SubSuite {
     /// List of tests belonging to sub-suite.
     pub test_list: Vec<String>,
 
+    /// Suite the generated suite is based off.
```
dbradf, line 168: Here are the new fields on the `SubSuite` struct that the config_writer is using. 

The `origin_suite` is which resmoke suite to base the configuration off and the `exclude_test_list` is used by _misc suites as what tests to exclude.

zamj (approved):

dbradf:
evergreen merge

PR #29, opened 2022-04-27
DAG-1804: Use task splitting fallback if too many missing stats

dbradf:


src/task_types/resmoke_tasks.rs
```diff
@@ -1187,4 +1251,18 @@ mod tests {
         assert_eq!(get_evg_fn_name(&commands[4]), Some("do multiversion setup"));
         assert_eq!(get_evg_fn_name(&commands[5]), Some("run test"));
     }
+
+    // percent_of_tests tests.
+    #[rstest]
+    #[case(100, 0.8, 80)]
+    #[case(100, 1.0, 100)]
```
MikhailShchatko, line 1258: I'm not sure if we should test this, but can we add some tests to see that results are rounded correctly?
e.g. `case(101, 0.8, ??)` or `case(99, 0.8, ??)`

src/task_types/resmoke_tasks.rs
```diff
@@ -991,6 +1020,41 @@ mod tests {
         assert!(suite_2.test_list.contains(&"test_5.js".to_string()));
     }
 
+    #[test]
+    fn test_split_task_without_missing_history_should_split_tasks_equally() {
```
MikhailShchatko, line 1024: I'm a bit confused, `without missing history` means that we have enough history data to split? If I'm not mistaken from the test code it seems that we don't have enough history data.

MikhailShchatko (commented):
Looks good % some nits

src/task_types/resmoke_tasks.rs
dbradf: That is a typo in the test name it should be `with missing history`.

MikhailShchatko (approved):
LGTM

dbradf:
evergreen merge

PR #32, opened 2022-05-16
DAG-1828: Support different exclude tags for last_lts and last_continuous

dbradf:


src/resmoke/resmoke_proxy.rs
```diff
@@ -133,8 +133,15 @@ impl TestDiscovery for ResmokeProxy {
 pub struct MultiversionConfig {
     /// Previous version of MongoDB to test against.
     pub last_versions: Vec<String>,
+
     /// Tags for required FCV version.
     pub requires_fcv_tag: String,
+
+    /// Tags for last LTS FCV versions.
+    pub requires_fcv_tag_lts: Option<String>,
```
dbradf, line 141: I'm planning to add these to the multiversion config output that resmoke provides. By making them `Option`, when they aren't provided, we will fallback to the `requires_fcv_tag` value that already exists. This will allow this change to be backwards-compatible.

zamj (approved):

dbradf:
evergreen merge

PR #34, opened 2022-06-06
DAG-1827 Pass through vars from "generate resmoke tasks" to "run generated tests" func

MikhailShchatko:


MikhailShchatko:
Tested it locally with adding new var to this task definition:
```
- <<: *gen_task_template
  name: retryable_writes_downgrade_last_continuous_gen
  tags: ["multiversion_passthrough", "multiversion", "no_version_combination"]
  commands:
  - func: "generate resmoke tasks"
    vars:
      multiversion_exclude_tags_version: last_continuous
```

Which was passed through:
```
    {
      "name": "retryable_writes_downgrade_last_continuous_0-enterprise",
      "commands": [
        {
          "func": "git get project no modules"
        },
        {
          "func": "add git tag"
        },
        {
          "func": "do setup"
        },
        {
          "func": "configure evergreen api credentials"
        },
        {
          "func": "do multiversion setup"
        },
        {
          "func": "run generated tests",
          "vars": {
            "suite": "generated_resmoke_config/retryable_writes_downgrade_last_continuous_0-enterprise.yml",
            "multiversion_exclude_tags_version": "last_continuous",
            "gen_task_config_location": "mongodb-mongo-master/d599853df5ccf96ef652a25d18c0d9da7f337d08/generate_tasks/generated-config-mongodb_mongo_master_d599853df5ccf96ef652a25d18c0d9da7f337d08.tgz",
            "resmoke_args": "--originSuite=retryable_writes_downgrade_last_continuous   --tagFile=generated_resmoke_config/multiversion_exclude_tags.yml --excludeWithAnyTags=multiversion_incompatible,backport_required_multiversion,retryable_writes_downgrade_last_continuous_backport_required_multiversion,requires_fcv_51,requires_fcv_52,requires_fcv_53,requires_fcv_60,requires_fcv_61",
            "require_multiversion_setup": true
          }
        }
      ],
      "depends_on": [
        {
          "name": "archive_dist_test_debug"
        }
      ]
    },
```

dbradf:
Can you fix the lint error?

dbradf:
> Can you fix the lint error?

Looking closer, it looks like the lint error is unrelated to your change. In CI we run against the latest version of the rust toolchain which is updated every 6 weeks. And new lint rules get added. It looks like a new lint rule is being triggered that is causing this failure.

In this case the rule checks to make sure that mutex locks are not held when await is called -- as this could cause a deadlock. In our case, this is happening near the end of the execution and by the time we get to this point in the code, nothing else will attempt to take the mutex, so it is not a problem. But it would be good to fix anyway, in case someone attempts to add more code that takes that mutex. 

The mutex is guarding the task definitions we generated. And we need to grab them to create the shrub configuration.

Here is a diff that should fix it:
```
diff --git a/src/lib.rs b/src/lib.rs
index 4c4b9a9..1fed832 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -242,11 +242,13 @@ pub async fn generate_configuration(
     let generated_build_variants =
         generate_tasks_service.generate_build_variants(generated_tasks.clone())?;
 
-    let generated_tasks = generated_tasks.lock().unwrap();
-    let task_defs: Vec<EvgTask> = generated_tasks
-        .values()
-        .flat_map(|g| g.sub_tasks())
-        .collect();
+    let task_defs: Vec<EvgTask> = {
+        let generated_tasks = generated_tasks.lock().unwrap();
+        generated_tasks
+            .values()
+            .flat_map(|g| g.sub_tasks())
+            .collect()
+    };
 
     let gen_evg_project = EvgProject {
         buildvariants: generated_build_variants.to_vec(),

```

MikhailShchatko:
@dbradf thanks for the fix and explanation.

dbradf (approved):

MikhailShchatko:
evergreen merge

PR #39, opened 2022-06-30
DAG-1879: Add support for generating burn_in_tests

dbradf:


dbradf:
This still needs more testing, but I wanted to get it out so people could start taking a look.

src/evergreen/evg_config.rs
```diff
@@ -44,8 +45,12 @@ impl EvgConfigService for EvgProjectConfig {
     }
 
     /// Get a map of task name to task definitions.
-    fn get_task_def_map(&self) -> HashMap<String, &EvgTask> {
-        self.evg_project.task_def_map()
+    fn get_task_def_map(&self) -> HashMap<String, EvgTask> {
```
dbradf, line 48: Since we will now be sharing this among threads, it was easier to take ownership of the `EvgTask` rather than having a reference to it.

src/resmoke/external_cmd.rs
```diff
@@ -0,0 +1,40 @@
...
+        .args(args)
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()?
+        .wait_with_output()?;
+
+    if !cmd.status.success() {
+        let error_message = String::from_utf8_lossy(&cmd.stderr).to_string();
+
+        event!(
```
dbradf, line 28: With this function we will log stderr output when we encounter an error trying to execute commands. We should be able to switch resmoke calls over to use this function and then they will have better error messages. I didn't add this here to keep this PR focused.

src/task_types/burn_in_tests.rs
```diff
@@ -0,0 +1,334 @@
...
+    ) -> EvgTask {
+        let origin_suite = suite_info.build_origin_suite(&params.suite_name);
+
+        let sub_suite = SubSuite {
+            index: Some(index),
+            name: suite_info.build_display_name(),
+            test_list: vec![test.to_string()],
+            exclude_test_list: None,
+            origin_suite: origin_suite.clone(),
+            mv_exclude_tags: suite_info.multiversion_tags.clone(),
```
MikhailShchatko, line 214: Did we do this thing with adding tags in the old burn in task generator?
If this is just keeping the same functionality, then it's fine. I just don't remember, if resmoke is able to exclude the test from running, if we pass the test path as an argument.

src/task_types/burn_in_tests.rs
```diff
@@ -0,0 +1,334 @@
...
+                    task_def,
+                    build_variant,
+                )?);
+            }
+        }
+
+        Ok(Box::new(GeneratedResmokeSuite {
+            task_name: "burn_in_tests".to_string(),
+            sub_suites,
+            use_large_distro: false,
```
MikhailShchatko, line 259: I assume `use_large_distro: false` is also keeping the same as it was before, right?

src/lib.rs
```diff
@@ -631,12 +680,48 @@ fn create_task_worker(
     })
 }
 
+/// Spawn a tokio task to perform the burn_in_test generation work.
+///
+/// # Arguments
+///
+/// * `deps` - Service dependencies.
+/// * `build_variant` - Build variant to query timing information from.
+/// * `generated_tasks` - Map to stored generated to in.
```
MikhailShchatko, line 689: [nit] Not getting this part `generated to in`

src/task_types/burn_in_tests.rs
```diff
@@ -0,0 +1,334 @@
...
+    /// Generate a burn_in_task for the given build variant.
+    ///
+    /// # Arguments
+    ///
+    /// * `build_variant` - Name of build variant to generate burn_in for.
+    /// * `task_map` - Map of task definitions found in the evergreen project configuration.
+    ///
+    /// # Returns
+    ///
+    /// A generated task for burn_in on the given build variant..
```
MikhailShchatko, line 34: [nit] Double dot at the end of line

MikhailShchatko (commented):
Just a couple of questions and some nits.

src/lib.rs
dbradf: fixed.

src/task_types/burn_in_tests.rs
dbradf: fixed.

src/task_types/burn_in_tests.rs
dbradf: I don't think burn_in has ever supported "use_large_distro", it just generate tasks on the default distro for the build variant. I believe there is a request on the backlog to start supporting large distros, but we would need to think through how best to do that.

src/task_types/burn_in_tests.rs
dbradf: This could be a problem that needs more investigation. I don't think burn_in has been running multiversion tests properly, so we likely haven't hit this before. But we should investigate how resmoke will behave. Ideally, resmoke would just no-op in this case, which is what I think we would want.

src/task_types/burn_in_tests.rs
MikhailShchatko: Got it.

src/task_types/burn_in_tests.rs
MikhailShchatko: I've checked `--excludeWithAnyTags` resmoke's flag, the resmoke no-oped and didn't run the test even if I passed it as an argument. So if `mv_exclude_tags` end up being in `--excludeWithAnyTags` list, then it should work.

MikhailShchatko:
evergreen merge

PR #41, opened 2022-07-13
DAG-1928: Add support for generating burn_in_tags

MikhailShchatko:
The original burn_in_tags python script: https://github.com/mongodb/mongo/blob/f4de1bd3b7de8b3e93d7d051cbfe8544df71112f/buildscripts/burn_in_tags.py

### DIFF
--- a/src/evergreen_names.rs
+++ b/src/evergreen_names.rs
@@ -34,6 +34,10 @@ pub const GENERATE_RESMOKE_TASKS: &str = "generate resmoke tasks";
 pub const GENERATOR_TASKS: &str = "generator_tasks";
 /// Name of burn_in_tests task.
 pub const BURN_IN_TESTS: &str = "burn_in_tests_gen";
+/// Name of burn_in_tags task.
+pub const BURN_IN_TAGS: &str = "burn_in_tags_gen";
+/// Name of compile task group.
+pub const COMPILE_TASK_GROUP_NAME: &str = "compile_and_archive_dist_test_TG";
 
 // Vars
 /// Variable that indicates a task is a fuzzer.
@@ -81,6 +85,10 @@ pub const MULTIVERSION_EXCLUDE_TAGS: &str = "multiversion_exclude_tags_version";
 // Build Variant expansions.
 /// Name of large distro for build variant.
 pub const LARGE_DISTRO_EXPANSION: &str = "large_distro_name";
+/// List of build variant names delimited by spaces to generate burn_in_tags for.
+pub const BURN_IN_TAG_BUILD_VARIANTS: &str = "burn_in_tag_buildvariants";
+/// Name of build variant to determine the timeouts for.
+pub const BURN_IN_BYPASS: &str = "burn_in_bypass";
 
 // Task Tags
 /// Tag to include multiversion setup is required.
@@ -99,3 +107,7 @@ pub const MULTIVERSION_EXCLUDE_TAGS_FILE: &str = "multiversion_exclude_tags.yml"
 pub const MULTIVERSION_LAST_LTS: &str = "last_lts";
 /// Name of last continuous configuration.
 pub const MULTIVERSION_LAST_CONTINUOUS: &str = "last_continuous";
+
+// Distros
+/// Name of compile task distro for burn_in_tags.
+pub const COMPILE_TASK_DISTRO: &str = "rhel80-xlarge";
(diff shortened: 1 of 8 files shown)


src/evergreen_names.rs
```diff
@@ -99,3 +107,7 @@ pub const MULTIVERSION_EXCLUDE_TAGS_FILE: &str = "multiversion_exclude_tags.yml"
 pub const MULTIVERSION_LAST_LTS: &str = "last_lts";
 /// Name of last continuous configuration.
 pub const MULTIVERSION_LAST_CONTINUOUS: &str = "last_continuous";
+
+// Distros
+/// Name of compile task distro for burn_in_tags.
+pub const COMPILE_TASK_DISTRO: &str = "rhel80-xlarge";
```
MikhailShchatko, line 113: I don't like this, but it was hardcoded this way in burn_in_tags.py and I didn't find a way to get this value from somewhere else.

We can add a mandatory buildvariant expansion for burn_in_tags, like we already have `burn_in_tag_buildvariants` expansion: https://github.com/mongodb/mongo/blob/master/etc/evergreen.yml#L1186
We can add `burn_in_tag_compile_distro`, but this will work only if we want to have them all to be the same distro. 

This is actually should be a map `burn_in_tag_build_variant_name: burn_in_tag_compile_distro`.

src/lib.rs
```diff
@@ -611,6 +658,23 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
             }
         }
 
+        for base_bv_name in burn_in_tag_build_variant_names {
+            let generated_tasks = generated_tasks.lock().unwrap();
+            let base_build_variant = build_variant_map.get(&base_bv_name).unwrap();
+            let run_build_variant_name = format!("{}-required", base_build_variant.name);
+            let task_name = format!("burn_in_tests-{}", run_build_variant_name);
```
zamj, line 665: Can you make this `burn_in_tests` a constant?

src/lib.rs
```diff
@@ -703,13 +770,15 @@ fn create_burn_in_worker(
 
     tokio::spawn(async move {
         let generated_task = burn_in_service
-            .generate_burn_in_suite(&build_variant.name, task_map)
+            .generate_burn_in_suite(&build_variant.name, &run_build_variant_name, task_map)
             .unwrap();
 
-        let task_name = format!("burn_in_tests-{}", build_variant.name);
+        let task_name = format!("burn_in_tests-{}", run_build_variant_name);
```
zamj, line 776: Same here, can you use a constant here?

src/task_types/burn_in_tests.rs
```diff
@@ -257,6 +314,63 @@ impl BurnInService for BurnInServiceImpl {
...
+    /// # Arguments
+    ///
+    /// * `base_build_variant` - Build variant to generate burn_in_tags build variant based on.
+    /// * `run_build_variant_name` - Build variant name to run burn_in_tests task on.
+    /// * `generated_task` - Generated burn_in_tests task.
+    ///
+    /// # Returns
+    ///
+    /// A generated burn_in_tags build variant based on another build variant.
+    fn generate_burn_in_tags_build_variant(
```
zamj, line 329: Can you add some tests for this?

zamj (commented):
Have you tested this against the current burn_in_tags script? Do we get the same output as the old script?

src/lib.rs
MikhailShchatko: Done.

src/lib.rs
```diff
@@ -703,13 +770,15 @@ fn create_burn_in_worker(
...
 
-        let mut generated_tasks = generated_tasks.lock().unwrap();
-        generated_tasks.insert(task_name, generated_task);
+        if !generated_task.sub_tasks().is_empty() {
+            let mut generated_tasks = generated_tasks.lock().unwrap();
+            generated_tasks.insert(task_name, generated_task);
+        }
```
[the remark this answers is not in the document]
MikhailShchatko: Done.

src/task_types/burn_in_tests.rs
MikhailShchatko: Done.

src/evergreen_names.rs
zamj: Can you cut a ticket to describe that issue and track the future work?

MikhailShchatko: Sure, will do after merging this.

zamj (approved):

MikhailShchatko:
evergreen merge

PR #42, opened 2022-07-19
DAG-1934 Add support for large distros in burn_in_tests

guoyr:
Ready for a first look. I will flesh out the unit test later today.

### DIFF
--- a/src/task_types/burn_in_tests.rs
+++ b/src/task_types/burn_in_tests.rs
@@ -1,6 +1,9 @@
 use anyhow::Result;
-use shrub_rs::models::task::EvgTask;
-use std::{collections::HashMap, sync::Arc};
+use shrub_rs::models::{task::EvgTask, task::TaskRef, variant::BuildVariant};
+use std::{
+    collections::{HashMap, HashSet},
+    sync::Arc,
+};
 
 use crate::{
     resmoke::burn_in_proxy::{BurnInDiscovery, DiscoveredTask},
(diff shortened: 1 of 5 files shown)


src/task_types/burn_in_tests.rs
```diff
@@ -220,41 +223,75 @@ impl BurnInServiceImpl {
...
+
+        for task_ref in variant_task_refs {
+            if task_ref.distros.is_some() {
+                large_distro_tasks.insert(task_ref.name.clone());
+            }
+        }
+
+        for discovered_task in discovered_tasks {
+            if large_distro_tasks.contains(&discovered_task.task_name) {
+                return true;
```
MikhailShchatko, line 253: So with this if any of the discovered tasks will have `use_large_distro: true` option, all burn_in sub tasks will run on a large distro, even if some of them are not running on large distro in a normal run.

I think if we would just default `use_large_distro` to `true` for all burn_in_tests, it will be very much the same and much simpler. Just as a quick fix to reduce complaints about burn_in_tests failures due to average time comparison with historic data.

To make those average test time checks in burn_in_tests more valuable I think we should move this [use_large_distro](https://github.com/mongodb/mongo-task-generator/blob/master/src/task_types/resmoke_tasks.rs#L231) from display_task to sub_suite level and set it in accordance to the original task definition.

zamj: I agree with Misha, I think we should this to be tracked at the sub_suite level.

I think a lot of the value from `burn_in` comes from being able to see at a glance if your changes have introduced an easily observable change in runtime. So, I think we should try to replicate the environment on a sub task by sub task basis if possible so that we're not accidentally hiding runtime increases by running things on a bigger, faster machine.

guoyr: Yeah valid points. Let me update the sub_suite part to accommodate task-level properties like distro sizes

guoyr:
Closing this out as I didn't get a chance to follow up.

PR #43, opened 2022-08-01
DAG-1940 Generate tasks separately for Windows, MacOS, Linux distro groups

MikhailShchatko:
After these changes a version generation for mongodb-mongo-master project has increased from ~337 sec to ~372 secs on my local machine:
```
{"timestamp":"2022-08-01T10:30:16.525079Z","level":"INFO","fields":{"message":"generation completed: 337 seconds"},"target":"mongo_task_generator"}
```

```
{"timestamp":"2022-08-01T10:39:08.927945Z","level":"INFO","fields":{"message":"generation completed: 372 seconds"},"target":"mongo_task_generator"}
```

Currently in evergreen it generates the version in ~143 seconds:
https://evergreen.mongodb.com/task_log_raw/mongodb_mongo_master_generate_tasks_for_version_version_gen_99286ff7b1f837df8449ef990b881c3ed1e3a64b_22_08_01_04_38_52/0?type=T#L915
```
{"timestamp":"2022-08-01T04:45:40.350646Z","level":"INFO","fields":{"message":"generation completed: 143 seconds"},"target":"mongo_task_generator"}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.5.0 - 2022-08-01
+
+* Generate tasks separately for Windows, MacOS, Linux distro groups.
+
 ## 0.4.7 - 2022-07-14
 
 * Add support for burn_in_tags generation.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1219,7 +1219,7 @@ dependencies = [
 
 [[package]]
 name = "mongo-task-generator"
-version = "0.4.7"
+version = "0.5.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mongo-task-generator"
-version = "0.4.7"
+version = "0.5.0"
 repository = "https://github.com/mongodb/mongo-task-generator"
 authors = ["Decision Automation Group <dev-prod-dag@mongodb.com>"]
 edition = "2018"
--- a/src/evergreen/evg_config_utils.rs
+++ b/src/evergreen/evg_config_utils.rs
@@ -1,4 +1,5 @@
 use std::collections::{HashMap, HashSet};
+use std::vec;
 
 use anyhow::{bail, Result};
 use lazy_static::lazy_static;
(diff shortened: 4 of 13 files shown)


zituo-jin:
Are we going to open another PR to fix uneven workloads? How does task splitting generate workloads with this change?

MikhailShchatko:
One of the reasons that tasks was split unevenly is that before these changes tasks were split based on historic runtime information from one of the build variants that runs the task.

It turned out that on some buildvariants some of the tests are running significantly longer than on other buildvariants, but with the different ratio than the other tests, e.g. runtimes can be like
test1.js - linux 10 secs, windows 12 secs, macos 14 secs
test2.js - linux 10 secs, windows 15 secs, macos 777 secs

So if we split task based on macos historic data, it will be unevenly split for linux and windows. Ideally we should split per each buildvariant historic data, but it will dramatically increase the version_gen task runtime. That's why we decided to split differently on major platforms.

I ran a patch build with these changes. The results are not perfect, but better than before:
https://evergreen.mongodb.com/task/mongodb_mongo_master_nightly_windows_display_aggregation_secondary_reads_patch_27c4d326ad61ffe9140127e368b348fbd32d9bca_62e8f1e532f4175239530161_22_08_02_09_44_27
https://evergreen.mongodb.com/task/mongodb_mongo_master_nightly_enterprise_windows_display_aggregation_secondary_reads_patch_27c4d326ad61ffe9140127e368b348fbd32d9bca_62e8f1e532f4175239530161_22_08_02_09_44_27
https://evergreen.mongodb.com/task/mongodb_mongo_master_nightly_enterprise_amazon2_arm64_display_aggregation_secondary_reads_patch_27c4d326ad61ffe9140127e368b348fbd32d9bca_62e8f1e532f4175239530161_22_08_02_09_44_27
https://evergreen.mongodb.com/task/mongodb_mongo_master_nightly_macos_arm64_display_aggregation_secondary_reads_patch_27c4d326ad61ffe9140127e368b348fbd32d9bca_62e8f1e532f4175239530161_22_08_02_09_44_27

To make it better we also need to update the splitting algorithm itself:
https://github.com/mongodb/mongo-task-generator/blob/0babc649532f5489fe3e7f24500c0364e099fdda/src/task_types/resmoke_tasks.rs#L434-L470

I didn't want to do this as part of this ticket. I will file another DAG ticket to update the splitting algorithm.

zituo-jin (approved):

MikhailShchatko:
evergreen merge

PR #44, opened 2022-08-11
DAG-1963 get the correct suite file in burn_in for non-generated tasks

guoyr:


src/task_types/burn_in_tests.rs
```diff
@@ -170,6 +170,7 @@ impl BurnInServiceImpl {
             let mut params = self
                 .config_extraction_service
                 .task_def_to_resmoke_params(task_def, false, None)?;
+
```
guoyr, line 173: will remove

src/evergreen/evg_config_utils.rs
```diff
@@ -308,11 +308,18 @@ impl EvgConfigUtils for EvgConfigUtilsImpl {
...
+        
+        if let Some(vars) = optional_vars {
+            if let Some(ParamValue::String(suite_var)) = vars.get("suite") {
+                &suite_var
+            } else {
+                generated_task_name
+            }
         } else {
-            remove_gen_suffix(&task.name)
+            generated_task_name
```
MikhailShchatko, line 322: Can we add some tests for these updates?

src/evergreen/evg_config_utils.rs
```diff
@@ -654,6 +661,42 @@ fn get_generate_resmoke_func(task: &EvgTask) -> Option<&FunctionCall> {
...
+    if let Some(Function(func)) = command {
+        if let Some(vars) = &func.vars {
+            return Some(vars)
+        } else {
+            None
+        }
+    } else {
+        None
+    }
+}
```
MikhailShchatko, line 698: And some tests for this as well

src/evergreen_names.rs
```diff
@@ -29,6 +29,9 @@ pub const ADD_GIT_TAG: &str = "add git tag";
 // Functions for generating tasks.
 pub const GENERATE_RESMOKE_TASKS: &str = "generate resmoke tasks";
 
+// Functions for running tests.
+pub const RUN_RESMOKE_TASK: &str = "run tests";
```
MikhailShchatko, line 33: Seems confusing. Maybe rename it to `RUN_RESMOKE_TESTS` or just `RUN_TESTS`?

MikhailShchatko (commented):
Looks great % some comments

src/evergreen/evg_config_utils.rs
guoyr: For this function, my thought was to have it be internal to this module. Since we don't need to mock out functions as much in this project (and rely instead more heavily on traits), the 2 tests added for `find_suite_name` should catch any visible changes to this function. Please feel free to push back

src/evergreen_names.rs
guoyr: Changed to `RUN_TEST_TESTS`. Adding some comments for posterity.

I'm a bit ambivalent about the naming. Conceptually, I think this function is both parallel to `_gen` tasks and downstream from them by being the generated task.

For the purposes of task generation logic in this codebase, including burn_in, we treat existing instances of this function as a part of the input for task generation, which means it's treated as an equal to the `_gen` task definitions. I therefore named it similar to `generate_resmoke_tasks` to avoid potential confusion of handling "generate_resmoke_tasks" and "run_tests" the same way.

src/evergreen/evg_config_utils.rs
zamj: That's fine with me

src/evergreen/evg_config_utils.rs
```diff
@@ -774,7 +833,27 @@ mod tests {
...
+        };
+        let evg_config_utils = EvgConfigUtilsImpl::new();
+
+        assert_eq!(evg_config_utils.find_suite_name(&evg_task), "my_task");
+    }
+
+    #[test]
+    fn test_find_suite_name_should_use_task_name_for_non_generated_task_if_no_var() {
+        let evg_task = EvgTask {
+            name: "my_task_gen".to_string(),
```
zamj, line 847: [nit] I think this should be `my_task` as it's not a gen task

zamj (approved):
LGTM % mod my comment

guoyr:
evergreen merge

PR #47, opened 2022-08-26
DAG-1992 Generate only burn_in tasks when --burn-in is passed

MikhailShchatko:
When `--burn-in` is passed, only burn_in tasks are generated. Otherwise all the other tasks are generated.

### DIFF
--- a/docs/generating_tasks.md
+++ b/docs/generating_tasks.md
@@ -160,11 +160,11 @@ the `"no_version_combinations"` tag should also be included.
 ### Burn in tests and burn in tags
 
 Newly added or modified tests might become flaky. In order to avoid that, those tests can be run
-continuously multiple times in a row to see, if the results are consistent. This process is called
+continuously multiple times in a row to see if the results are consistent. This process is called
 burn-in.
 
 `burn_in_tests_gen` task is used to generate burn-in tasks on the same buildvariant the task is
-added to. [Example](https://github.com/mongodb/mongo/blob/81c41bdfdc56f05973fae70e80e80919f18f50c9/etc/evergreen_yml_components/definitions.yml#L3252-L3256)
+added to. The [example](https://github.com/mongodb/mongo/blob/81c41bdfdc56f05973fae70e80e80919f18f50c9/etc/evergreen_yml_components/definitions.yml#L3252-L3256)
 of task configuration:
 
 ```yaml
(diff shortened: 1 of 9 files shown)


MikhailShchatko:
Example of patch where burn_in and other tasks are generated separately: https://spruce.mongodb.com/version/6308dc350305b95cf9de852b

src/lib.rs
```diff
@@ -467,8 +469,8 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
             for task in &build_variant.tasks {
                 // Burn in tasks could be different for each build variant, so we will always
                 // handle them.
-                if task.name == BURN_IN_TESTS {
-                    if self.gen_burn_in {
+                if self.gen_burn_in {
```
zamj, line 472: Can you add some tests for this logic to make sure we're skipping the tasks when we should be?

zamj (approved):
Overall LGTM after adding tests

zamj:
We should update the docs as well for this option

src/lib.rs
MikhailShchatko: This particular function seems hard to test or split. I've added end-to-end test for this logic.

Also added a comment to make it more clear:
https://github.com/mongodb/mongo-task-generator/pull/47/files#diff-7ff815b68943f08bef79f083c09e55ca8c18db8e25033034b12fa31f1d43dab4R64-R65

Let me know if you have better ideas.

docs/generating_tasks.md
```diff
@@ -157,6 +157,49 @@ of the task definition. When this tag is present, both the extra setup steps and
...
+- <<: *gen_task_template
+  name: burn_in_tests_gen
+  tags: []
+  commands:
+  - func: "generate resmoke tasks"
+```
+
+`burn_in_tags_gen` task is used to generate separate burn-in buildvariants. This way we can add the
+task to the required buildvariant, that will generate another required buildvariants on the fly
+based on another non-required buildvariants with only necessary burn-in tasks.
```
zamj, line 180: [nit] This way we can burn_in on the the requested buildvariant as well other, additional variants to ensure there is no difference between variants.

docs/generating_tasks.md
```diff
@@ -157,6 +157,49 @@ of the task definition. When this tag is present, both the extra setup steps and
 of multiversion sub-tasks will be preformed. In order to only perform the extra setup steps
 the `"no_version_combinations"` tag should also be included.
 
+### Burn in tests and burn in tags
+
+Newly added or modified tests might become flaky. In order to avoid that, those tests can be run
+continuously multiple times in a row to see, if the results are consistent. This process is called
```
zamj, line 163: [nit] Can you remove this comma

docs/generating_tasks.md
```diff
@@ -157,6 +157,49 @@ of the task definition. When this tag is present, both the extra setup steps and
...
+```yaml
+- <<: *gen_task_template
+  name: burn_in_tags_gen
+  tags: []
+  commands:
+  - func: "generate resmoke tasks"
+```
+
+`burn_in_tag_buildvariants` buildvariant expansion is used to configure base buildvariant names.
+Base buildvariant names should be delimited by spaced. [Example](https://github.com/mongodb/mongo/blob/81c41bdfdc56f05973fae70e80e80919f18f50c9/etc/evergreen.yml#L1257)
```
zamj, line 194: [nit] spaces

zamj (approved):
LGMT mod my comments

docs/generating_tasks.md
MikhailShchatko: updated

docs/generating_tasks.md
MikhailShchatko: updated

docs/generating_tasks.md
MikhailShchatko: updated

MikhailShchatko:
evergreen merge

PR #48, opened 2022-08-29
DAG-1950: Make burn_in_tags compile task group and distro name configurable

zamj:
Patch:https://spruce.mongodb.com/version/630e63453066156e2ff2b9ee/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/evergreen_names.rs
+++ b/src/evergreen_names.rs
@@ -39,8 +39,6 @@ pub const GENERATOR_TASKS: &str = "generator_tasks";
 pub const BURN_IN_TESTS: &str = "burn_in_tests_gen";
 /// Name of burn_in_tags task.
 pub const BURN_IN_TAGS: &str = "burn_in_tags_gen";
-/// Name of compile task group.
-pub const COMPILE_TASK_GROUP_NAME: &str = "compile_and_archive_dist_test_TG";
 
 // Vars
 /// Variable that indicates a task is a fuzzer.
@@ -90,6 +88,10 @@ pub const MULTIVERSION_EXCLUDE_TAGS: &str = "multiversion_exclude_tags_version";
 pub const LARGE_DISTRO_EXPANSION: &str = "large_distro_name";
 /// List of build variant names delimited by spaces to generate burn_in_tags for.
 pub const BURN_IN_TAG_BUILD_VARIANTS: &str = "burn_in_tag_buildvariants";
+/// The distro to use when compiling burn_in_tags.
+pub const BURN_IN_TAG_COMPILE_DISTRO: &str = "burn_in_tag_compile_distro";
+/// The name to give the task group for compiling.
+pub const BURN_IN_TAG_COMPILE_TASK_GROUP_NAME: &str = "burn_in_tag_compile_task_group_name";
 /// Name of build variant to determine the timeouts for.
 pub const BURN_IN_BYPASS: &str = "burn_in_bypass";
 
@@ -110,10 +112,6 @@ pub const MULTIVERSION_EXCLUDE_TAGS_FILE: &str = "multiversion_exclude_tags.yml"
 pub const MULTIVERSION_LAST_LTS: &str = "last_lts";
 /// Name of last continuous configuration.
 pub const MULTIVERSION_LAST_CONTINUOUS: &str = "last_continuous";
-
-// Distros
-/// Name of compile task distro for burn_in_tags.
-pub const COMPILE_TASK_DISTRO: &str = "rhel80-xlarge";
 
 // Distro group names
 /// Windows distro group name.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -334,6 +334,24 @@ trait GenerateTasksService: Sync + Send {
         generated_tasks: Arc<Mutex<GenTaskCollection>>,
     ) -> Result<Vec<BuildVariant>>;
 
+    /// Generate the burn_in build variant information for a build variant.
+    ///
+    /// # Arguments
+    ///
+    /// * `burn_in_tag_build_variant_info` - A map of burn_in build variants to config information about them.
+    /// * `build_variant` - The original build variant to generate burn_in information from.
+    /// * `build_variant_map` - A map of build variant names to their definitions.
+    ///
+    /// # Returns
+    ///
+    /// Nothing, modifies the burn_in_tag_build_variant_info with new values.
+    fn generate_burn_in_build_variant_info(
+        &self,
+        burn_in_tag_build_variant_info: &mut HashMap<String, BurnInTagBuildVariantInfo>,
+        build_variant: &BuildVariant,
+        build_variant_map: &HashMap<String, &BuildVariant>,
+    );
+
     /// Generate a task for the given task definition.
     ///
     /// # Arguments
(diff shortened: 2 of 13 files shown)


src/evergreen_names.rs
```diff
@@ -90,6 +90,10 @@ pub const MULTIVERSION_EXCLUDE_TAGS: &str = "multiversion_exclude_tags_version";
 pub const LARGE_DISTRO_EXPANSION: &str = "large_distro_name";
 /// List of build variant names delimited by spaces to generate burn_in_tags for.
 pub const BURN_IN_TAG_BUILD_VARIANTS: &str = "burn_in_tag_buildvariants";
+/// List of build variant names delimited by spaces to generate burn_in_tags for.
+pub const BURN_IN_TAG_COMPILE_DISTRO: &str = "burn_in_tag_compile_distro";
+/// List of build variant names delimited by spaces to generate burn_in_tags for.
+pub const BURN_IN_TAG_COMPILE_TASK_GROUP_NAME: &str = "burn_in_tag_compile_task_group_name";
```
MikhailShchatko, line 96: Docstrings need to be updated.

src/lib.rs
```diff
@@ -680,13 +682,26 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
...
+                .evg_config_utils
+                .lookup_build_variant_expansion(BURN_IN_TAG_COMPILE_DISTRO, base_build_variant)
+                .unwrap_or_else(|| DEFAULT_BURN_IN_COMPILE_TASK_DISTRO.to_string());
+            let compile_task_group_name = self
+                .evg_config_utils
+                .lookup_build_variant_expansion(
+                    BURN_IN_TAG_COMPILE_TASK_GROUP_NAME,
+                    base_build_variant,
+                )
+                .unwrap_or_else(|| DEFAULT_BURN_IN_COMPILE_TASK_GROUP_NAME.to_string());
```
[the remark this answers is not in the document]
MikhailShchatko: The current approach LGTM

MikhailShchatko (approved):

zamj:
@MikhailShchatko Here's a patch with the new approach, it seems to have picked up the task name. It failed as it wasn't a valid name. 
https://spruce.mongodb.com/version/630fa54aa4cf474cd9b687c2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

zamj:
@MikhailShchatko I made some updates to this overall. The issue seems to have been that burn_in_variants could be defined in multiple different places potentially so distros and task_group_names were overwriting each other.

Patch: https://spruce.mongodb.com/task/mongodb_mongo_master_generate_tasks_for_version_version_gen_patch_f3b689770ec69bd9ac089ec4b96a678bb0d565b9_63111515850e61077e0d8cf3_22_09_01_20_25_14/files?execution=0

src/lib.rs
```diff
@@ -618,13 +626,38 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
...
+                            )
+                            .unwrap_or_else(|| DEFAULT_BURN_IN_COMPILE_TASK_GROUP_NAME.to_string());
+                        for variant in burn_in_tag_build_variants {
+                            let bv_info = burn_in_tag_build_variant_info.entry(variant).or_insert(
+                                BurnInTagBuildVariantInfo {
+                                    compile_distros: HashSet::from([compile_distro.to_string()]),
+                                    compile_task_group_name: compile_task_group_name.clone(),
+                                },
+                            );
+                            bv_info.compile_distros.insert(compile_distro.to_string());
```
MikhailShchatko, line 656: Do we want to apply the same logic here and add only non-default distros here as below?

MikhailShchatko (approved):
That's annoying, but LGTM as I don't see any better way to handle this.

src/lib.rs
zamj: I don't think we can avoid adding the default here like below. We'll always need at least one distro to send to Evergreen and if that distro isn't set in the Evergreen yaml, we'll need to add the default here.

What do you think?

MikhailShchatko: I was thinking that at least one distro will be already added before this line. Does it look right?

zamj: Yes, there should at least be one distro by the time this line hits. That distro could be the default distro though. I could add some logic here that, if the only distro is the default one, we replace the default with the non default distro. I think that might not be the best case though as theoretically someone could define the default distro as part of the evergreen yaml and we would be removing it.

The only time this should be a problem is if multiple different variants define the same burn_in_tag_build_variants. Which honestly seems like a bit of a strange case in my mind.
https://github.com/10gen/mongo/blob/master/etc/evergreen.yml#L1257
https://github.com/10gen/mongo/blob/master/etc/evergreen.yml#L1594
Here are the two places where we're currently defining the same burn_in variants. I think it would be fine to remove one?

MikhailShchatko: I think we should allow to have the same `burn_in_tag_buildvariants` defined in different places.

My understanding of burn_in_tags is that the user is just choosing to run all the tasks on some buildvariant and it will seamlessly generate burn_in_tags buildvariants. I think the majority of users should not be bothered selecting specifically `burn_in_tags_gen` task. So with the current evergreen configuration it will happen on `enterprise-rhel-80-64-bit-dynamic-required` and `enterprise-rhel-80-64-bit-dynamic-classic-engine` buildvariants and those generated buildvariants are happen to be the same.

Here is what I suggest:
1. Let's remove default values. We will explicitly define them for existing tasks and it will be a good example of how to configure some new tasks. 
2. Enforce consistent configuration. If there are several definitions of `burn_in_tag_buildvariants`, error out when the values of `burn_in_tag_compile_distro` and `burn_in_tag_compile_task_group_name` are different across those several definitions.

Let me know what you think.

MikhailShchatko:
I decided to merge https://github.com/mongodb/mongo-task-generator/commit/0c908229187f81c4fc9f0e6381729bd4f235aca4.
Could you please rebase your changes on top of it and update the docs with a new configuration [here](https://github.com/mongodb/mongo-task-generator/blob/0c908229187f81c4fc9f0e6381729bd4f235aca4/docs/generating_tasks.md?plain=1#L193-L199)?

src/lib.rs
zamj: That sounds like a good path forward to me

zamj:
Patches:
[Without distro](https://evergreen.mongodb.com/lobster/evergreen/task/mongodb_mongo_master_generate_tasks_for_version_version_gen_patch_f3b689770ec69bd9ac089ec4b96a678bb0d565b9_6317cf993e8e86303d9409f1_22_09_06_22_54_52/0/task#bookmarks=0%2C691&l=1)

src/lib.rs
```diff
@@ -622,13 +630,54 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
...
-                                    BURN_IN_TAG_BUILD_VARIANTS,
-                                    build_variant,
-                                ),
-                        );
+                        let burn_in_tag_build_variants = self
+                            .evg_config_utils
+                            .lookup_and_split_by_whitespace_build_variant_expansion(
+                                BURN_IN_TAG_BUILD_VARIANTS,
+                                build_variant,
+                            );
```
MikhailShchatko, line 638: Can we also check that `burn_in_tag_build_variants` vector is not empty? That should mean that `BURN_IN_TAG_BUILD_VARIANTS` expansion is absent or empty.

And sorry, I just remembered, but we also want to check that those values from `BURN_IN_TAG_BUILD_VARIANTS` expansion exist in `build_variant_map`. We should also error out in this case.

src/lib.rs
```diff
@@ -622,13 +630,54 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
...
+                                    build_variant.name, variant, BURN_IN_TAG_COMPILE_DISTRO
+                                )
+                            }
+                            if bv_info.compile_task_group_name != compile_task_group_name {
+                                panic!(
+                                    "`{}` is trying to set a different compile task group name than already exists for {}. Check the {} expansions in your config.",
+                                    build_variant.name, variant, BURN_IN_TAG_COMPILE_TASK_GROUP_NAME
+                                )
+                            }
+                        }
```
MikhailShchatko, line 680: Is it possible to extract this part to a separate function? `generate_build_variants` is getting very huge, I wonder if we can make it somewhat better.

MikhailShchatko (commented):
Looks great % some more follow up suggestions.

src/lib.rs
```diff
@@ -622,13 +630,54 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
...
+                            });
+                        let compile_task_group_name = self
+                            .evg_config_utils
+                            .lookup_build_variant_expansion(
+                                BURN_IN_TAG_COMPILE_TASK_GROUP_NAME,
+                                build_variant,
+                            ).unwrap_or_else(|| {
+                                panic!(
+                                    "`{}` is missing the burn_in_tag compile task group name expansion. Set the '{}' expansion in your project's config to continue.",
+                                    build_variant.name, BURN_IN_TAG_COMPILE_TASK_GROUP_NAME
```
MikhailShchatko, line 658: Can we update this message and the one above?
```
"`{}` build variant is missing the `{}` expansion to run `{}`. Set the expansion in your project's config to continue.",
build_variant.name, BURN_IN_TAG_COMPILE_TASK_GROUP_NAME, BURN_IN_TAGS
```

src/lib.rs
```diff
@@ -622,13 +630,54 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
...
+                            let bv_info = burn_in_tag_build_variant_info
+                                .entry(variant.clone())
+                                .or_insert(BurnInTagBuildVariantInfo {
+                                    compile_distro: compile_distro.clone(),
+                                    compile_task_group_name: compile_task_group_name.clone(),
+                                });
+                            if bv_info.compile_distro != compile_distro.clone() {
+                                panic!(
+                                    "`{}` is trying to set a different compile distro than already exists for {}. Check the {} expansions in your config.",
+                                    build_variant.name, variant, BURN_IN_TAG_COMPILE_DISTRO
```
MikhailShchatko, line 671: [nit]
```
"`{}` is trying to set a different compile distro than already exists for `{}`. Check the `{}` expansions in your config."
```

src/lib.rs
```diff
@@ -622,13 +630,54 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
...
+                                )
+                            }
+                            if bv_info.compile_task_group_name != compile_task_group_name {
+                                panic!(
+                                    "`{}` is trying to set a different compile task group name than already exists for {}. Check the {} expansions in your config.",
+                                    build_variant.name, variant, BURN_IN_TAG_COMPILE_TASK_GROUP_NAME
```
MikhailShchatko, line 677: [nit] 
```
"`{}` is trying to set a different compile task group name than already exists for `{}`. Check the `{}` expansions in your config."
```

MikhailShchatko (commented):
And some more suggestions.

MikhailShchatko (approved):
LGTM

zamj:
evergreen merge

PR #51, opened 2022-10-10
DAG-2208 update burn_in_tags to support existing compile tasks

guoyr:
Putting this piece for context as the downstream user of https://github.com/evergreen-ci/shrub-rs/pull/1


src/lib.rs
```diff
@@ -672,21 +668,21 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
...
             if bv_info.compile_task_group_name != compile_task_group_name {
                 panic!(
                     "`{}` is trying to set a different compile task group name than already exists for `{}`. Check the `{}` expansions in your config.",
                 build_variant.name, variant, BURN_IN_TAG_COMPILE_TASK_GROUP_NAME
             )
             }
+            if bv_info.compile_variant != compile_variant {
+                panic!(
+                    "`{}` is trying to set a different compile variant than already exists for `{}`. Check the `{}` expansions in your config.",
+                    build_variant.name, variant, COMPILE_VARIANT
```
MikhailShchatko, line 683: Could there be cases where we would want to have a dedicated `burn_in_tag_compile_variant` expansion instead of using a single `compile_variant` expansion for the current build variant and also burn_in_tag build variants it produces?

I assume it should be easier to understand how to configure burn_in_tags or what is wrong with configuration with dedicated expansions.

guoyr: Good question Misha. For this specific option, there's always been a desire to remove the dedicated compile variant for burn in, which is done as a part of this ticket. 

Longer-term, the plan is to move the really non-deterministic testing to mainline only so burn in doesn't become the scapegoat for all flaky test issues.

MikhailShchatko: All right, sounds good to me!

zamj:
@guoyr Overall, this LGTM! I'm curious to see what this looks like, do you have a patch handy where you ran this?

guoyr:
Sadly there's an issue with the Evergreen API that prevents burn_in_tags from activating generated tasks. I created https://jira.mongodb.org/browse/EVG-18112 and https://jira.mongodb.org/browse/DAG-2233 to track the followup work.

The burn_in_tags variants will be statically defined for now in evergreen.yml

guoyr:
@zamj merging this for now so I can more quickly bisect to the change that causes EVG-18112.

guoyr:
evergreen merge

PR #52, opened 2022-10-24
DAG-2208 uncomment workaround for EVG issue

guoyr:
It turned out that only the `build_by_id()` EVG API was broken. `tasks_by_build()` worked well as an alternative so I'm uncommenting the temporary excluded code from #51.

One other change is to add the `compile_variant` expansion that I had missed in the original PR.

I'm going to skip the line and merge this given the code was a part of #51. Please feel free to revert/amend if anything doesn't look right.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.6.5 - 2022-10-23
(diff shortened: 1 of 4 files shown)


guoyr:
evergreen merge

guoyr:
evergreen merge

PR #54, opened 2022-11-18
DAG-2264 Switch to using evergreen test stats from S3

MikhailShchatko:
Done based on https://github.com/mongodb/mongo-task-generator/pull/53

Some issues getting data from S3:
- Data is not available: [evergreen log](https://evergreen.mongodb.com/task_log_raw/mongodb_mongo_master_generate_tasks_for_version_version_gen_patch_5c0652aac3cb161cb8d0bcf0d80cdbb084a74d0c_6377aaee1e2d17611ac9ec38_22_11_18_15_55_44/0?type=T#L907) - [what S3 returns](https://mongo-test-stats.s3.amazonaws.com/mongodb-mongo-master/enterprise-rhel-80-64-bit-future-git-tag-multiversion/sharded_collections_jscore_multiversion)
- Sometimes getting [this error](https://evergreen.mongodb.com/task_log_raw/mongodb_mongo_master_generate_tasks_for_version_version_gen_patch_5c0652aac3cb161cb8d0bcf0d80cdbb084a74d0c_6377aaee1e2d17611ac9ec38_22_11_18_15_55_44/0?type=T#L1247) even with retries, or maybe retry doesn't work in such cases
```
"error":"Request error: error sending request for url (https://mongo-test-stats.s3.amazonaws.com/mongodb-mongo-master/enterprise-rhel-80-64-bit-dynamic-all-feature-flags-required/sharding_max_mirroring): channel closed"
```

### DIFF
--- a/src/evergreen/evg_config.rs
+++ b/src/evergreen/evg_config.rs
@@ -109,7 +109,7 @@ impl EvgConfigService for EvgProjectConfig {
 /// Shrub representation of evergreen configuration.
 fn get_project_config(location: &Path) -> Result<EvgProject> {
     let evg_config_yaml = Command::new("evergreen")
-        .args(&["evaluate", location.to_str().unwrap()])
+        .args(["evaluate", location.to_str().unwrap()])
         .output()?;
     Ok(EvgProject::from_yaml_str(std::str::from_utf8(&evg_config_yaml.stdout)?).unwrap())
 }


julianedwards:
So I am not sure why S3 gives an accessed denied error instead of a 404 from the first issue you saw, but the data does not exist. I checked the test results table and cross-checked with the existing evergreen test stats endpoint to verify that no test results/stats data for `mongodb-mongo-master/enterprise-rhel-80-64-bit-future-git-tag-multiversion/sharded_collections_jscore_multiversion` exists between 2022-10-03 and 2022-10-17.

Evergreen test stats endpoint:
https://evergreen.mongodb.com/rest/v2/projects/mongodb-mongo-master/test_stats?after_date=2022-11-03&before_date=2022-11-18&group_num_days=14&requesters=mainline&tasks=sharded_collections_jscore_multiversion&variants=enterprise-rhel-80-64-bit-future-git-tag-multiversion

As for the channel closed error, I am not sure what that is about. Could it be something rust specific?

julianedwards:
Ah I think I know why you get accessed denied—the bucket by default is private. We set the ACL policy (public-read) on each file we write rather than the bucket itself, since that file does not exist it uses the bucket ACL policy and refuses to give you any more information except for "access denied".

julianedwards (approved):

zamj:
> Ah I think I know why you get accessed denied—the bucket by default is private. We set the ACL policy (public-read) on each file we write rather than the bucket itself, since that file does not exist it uses the bucket ACL policy and refuses to give you any more information except for "access denied".

Do we need to add any additional handling for this access denied case?

julianedwards:
> Do we need to add any additional handling for this access denied case?

I don't think so. You can change the bucket policy to `public-read`, but it won't make a difference in the behavior of the task generator tool because it falls back to naive task splitting either way, whether it gets a 404 or 403.

MikhailShchatko:
> > Ah I think I know why you get accessed denied—the bucket by default is private. We set the ACL policy (public-read) on each file we write rather than the bucket itself, since that file does not exist it uses the bucket ACL policy and refuses to give you any more information except for "access denied".
> 
> Do we need to add any additional handling for this access denied case?

If the data doesn't exist in evergreen then I think we are fine. We would not get that data from evergreen api anyways.

zamj (approved):

PR #55, opened 2022-12-13
DAG-2284 Support burn_in equivalent for tasks generation

MikhailShchatko:
A new burn_in for tasks is using `burn_in_task_names` build variant expansion to generate burn_in_tasks for. It is expected to run only in patches. The existing burn_in's effectively do their jobs only in patches, since in the versions on the waterfall there is no patch part and burn_in never finds anything to run. Therefore I think it is reasonable to switch all burn_in related tasks to run only in patches.

A new burn_in for tasks does not support task splitting and generates 10 (hard coded value) copies of the task that should run a full resmoke suite with all tests in each copy of the task.

This PR also includes `use_large_distro` support for burn_in. Previously burn_in did not take in account this configuration and could generate tasks to run on a smaller distro than it will run in a normal mode, which resulted in slower runs. I think this is important feature that is necessary for burn_in_tasks to get actual value from it.

Evergreen patch with triggered burn_in:
https://spruce.mongodb.com/version/6398984a9ccd4e45e88eb334
Evergreen patch with required builds:
https://spruce.mongodb.com/version/6398987aa4cf471e3e8da014


src/task_types/burn_in_tests.rs
```diff
@@ -231,15 +332,14 @@ impl BurnInServiceImpl {
...
         suite_info: &BurnInSuiteInfo,
-    ) -> EvgTask {
+    ) -> GeneratedSubTask {
         let origin_suite = suite_info.build_origin_suite(&params.suite_name);
 
         let sub_suite = SubSuite {
             index,
             name: suite_info.build_display_name(),
-            test_list: vec![test.to_string()],
+            test_list: vec![],
```
zamj, line 342: Don't we need to know the test name here? Will this break burn_in tests?

zamj (changes requested):
Nice job on the large distro support. Can you add some additional testing in here for the new code, such as the task generation? e2e tests would also be great. 

Also, can you create and link a patch using the new burn_in_tasks feature?

src/task_types/burn_in_tests.rs
MikhailShchatko: We don't need it here. We use resmoke invocation to pass in the test name [here](https://github.com/mongodb/mongo-task-generator/blob/2f23b6ff9b849763476baac9b4207a31c2498962/src/task_types/burn_in_tests.rs#L395).

As for this `test_list`, it is needed to write a physical suite yaml file. We use it during regular task splitting and never used it for burn_in_tests.

MikhailShchatko:
@zamj added more tests.

I've linked a patch above, but in fact it is hard to spot `burn_in_tasks` among a bunch of `burn_in_tests`, so here is the link to the generated task:
https://spruce.mongodb.com/task/mongodb_mongo_master_enterprise_rhel_80_64_bit_dynamic_display_burn_in_tasks_patch_aa37b60da0b57a84c210c20d0433a8ac7e3133a2_6398984a9ccd4e45e88eb334_22_12_13_15_20_56

zamj:
> @zamj added more tests.
> 
> I've linked a patch above, but in fact it is hard to spot `burn_in_tasks` among a bunch of `burn_in_tests`, so here is the link to the generated task: https://spruce.mongodb.com/task/mongodb_mongo_master_enterprise_rhel_80_64_bit_dynamic_display_burn_in_tasks_patch_aa37b60da0b57a84c210c20d0433a8ac7e3133a2_6398984a9ccd4e45e88eb334_22_12_13_15_20_56

Patch looks great!

I think it's fine if we restrict people to only "burning_in" on a single task at a time. Some tasks can be extremely massive and I'd like to try guard against someone accidentally creating an extremely massive patch full of very long running tasks.

docs/generating_tasks.md
```diff
@@ -200,6 +200,24 @@ burn_in_tag_compile_distro: rhel80-large
...
+  tags: []
+  commands:
+  - func: "generate resmoke tasks"
+```
+
+`burn_in_task_name` buildvariant expansion is used to configure which task to burn-in. The
+example of `burn_in_task_name` buildvariant expansion:
+
+```yaml
+burn_in_task_name: replica_sets_jscore_passthrough_gen
```
zamj, line 218: `_gen` tasks are a bit scary to me for us to support as part of `burn_in_tasks`. They're usually very large and are meant to be split up by the generator. I don't think we should split up these tasks as part of `burn_in_tasks` as that will have us wind up creating 50 different tasks in Evergreen.

Can we not allow any tasks that end in `_gen`? It seems likely that they would time out anyway for the most part as a lot of tasks are split up to avoid running forever in patches. Or maybe we should add this point as a warning in our docs to users. Do you have any thoughts @MikhailShchatko ?

MikhailShchatko: I think we can add a warning message and something in docs, that some tasks may time out, since task splitting is not supported in burn-in tasks. @zamj what do you think?

zamj: That sounds good to me

MikhailShchatko: Added information in docs.

Now I think it's not very useful to print a message by task generator since it's not very likely to me that somebody will look at `version_burn_in_gen` logs when their tasks will be timing out.

I think it's better to add a comment in mongo's `evergreen.yml` above `burn_in_task_name` expansion with the message.

@zamj let me know what you think?

zamj: I think that sounds good to me, I agree it's likely more visible there.

zamj (approved):

MikhailShchatko:
evergreen merge

PR #59, opened 2023-04-04
DAG-2546 Improve UX of generated multiversion tasks

trahman1318:
PB: https://spruce.mongodb.com/version/644062ad61837dc8a953c172/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
Corresponding Mongo PR: https://github.com/10gen/mongo/pull/11789

TODO:
- Update docs
- Confirm that everything from before & after is identical. I've done this for the most part, but need to remember to do this before merge.


Cargo.toml
```diff
@@ -25,7 +25,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 serde_yaml = "0.8"
 shellexpand = "2.1"
-shrub-rs = "0.5.1"
+shrub-rs = { path = "../shrub-rs-0.5.1" }
```
trahman1318, line 28: I needed to make some very small changes here as well, but I did not yet receive access to this repo via MANA. Will follow up with DAG.

src/evergreen/evg_config_utils.rs
```diff
@@ -6,6 +6,7 @@ use lazy_static::lazy_static;
 use regex::Regex;
 use shrub_rs::models::commands::EvgCommand::Function;
 use shrub_rs::models::params::ParamValue;
+use shrub_rs::models::task::MultiversionGenerateTaskConfig;
```
trahman1318, line 9: A new struct in the `shrub-rs` repo.

src/task_types/burn_in_tests.rs
trahman1318, line 1: I haven't updated this file properly yet, but this should be straightforward.

src/task_types/resmoke_config_writer.rs
trahman1318, line 1: Pre-emptively modified some tests here -- will fix later. No need to take a look.

trahman1318 (commented):
This isn't really ready for review yet. Just wanted to show the current progress.

trahman1318:
With these changes, Task Generation now succeeds with the new evergreen/multiversion API changes (https://github.com/10gen/mongo/pull/11789). That means the code here works, we just need to verify that test coverage is not lost. I have done a check by `ls`ing the generated files from the current and future task generator, which looks identical, but I need to look at the actual diff to confirm (need to get rid of some of the randomization to do this, which is why I havent done this yet).

TODO:
(This is more for me to keep track of what needs to be done)
- Check the actual diff & make sure things look very close to the same -- especially for `evergreen_config.json`
- Fix up burn_in_tests.rs (should be the same as what was done for `resmoke_tasks.rs` & `fuzzer_tasks.rs`
- Fix any broken tests & write any new tests (probably a couple unit tests & maybe an integration test)
- Update docstrings/docs
- Remove dead code from the now outdated `MultiversionService`

src/evergreen/evg_config_utils.rs
```diff
@@ -370,7 +411,7 @@ impl EvgConfigUtils for EvgConfigUtilsImpl {
     ///
     /// Value of given variable in the 'generate resmoke task' vars.
     fn get_gen_task_var<'a>(&self, task: &'a EvgTask, var: &str) -> Option<&'a str> {
-        let generate_func = get_generate_resmoke_func(task);
+        let generate_func = get_func_by_name(task, GENERATE_RESMOKE_TASKS);
```
trahman1318, line 414: I can probably use my new `get_func_vars_by_name` function from below here. Will try this again once other things are in a better state.

src/task_types/burn_in_tests.rs
```diff
@@ -210,7 +210,7 @@ impl BurnInServiceImpl {
                 .task_def_to_resmoke_params(task_def, false, None)?;
             update_resmoke_params_for_burn_in(&mut params, test);
 
-            if params.generate_multiversion_combos {
+            if params.require_multiversion_generate_tasks {
```
trahman1318, line 213:
```rust
for (old_version, version_combination) in self
    .multiversion_service
    .multiversion_iter(&params.suite_name)?
```
Need to fix that part up simliar to `fuzzer_tasks.rs` & `resmoke_tasks.rs`

src/task_types/fuzzer_tasks.rs
```diff
@@ -216,27 +217,24 @@ impl GenFuzzerService for GenFuzzerServiceImpl {
         let task_name = &params.task_name;
         let mut sub_tasks: Vec<EvgTask> = vec![];
         if params.is_multiversion() {
+            if params.multiversion_generate_tasks.is_empty() {
```
trahman1318, line 220: This works & is okay, but I think the more rust idiomatic way to do this is to make
```rust
pub multiversion_generate_tasks: Vec<MultiversionGenerateTaskConfig>,
```
into
```rust
pub multiversion_generate_tasks: Option<Vec<MultiversionGenerateTaskConfig>>,
```
And then unwrap the variable so the panic is built-in.

src/task_types/fuzzer_tasks.rs
```diff
@@ -216,27 +217,24 @@ impl GenFuzzerService for GenFuzzerServiceImpl {
...
-                    build_name(&params.task_name, &old_version, &version_combination);
-                let base_suite_name = build_name(&params.suite, &old_version, &version_combination);
-
+            for multiversion_task in &params.multiversion_generate_tasks {
                 sub_tasks.extend(
                     (0..params.num_tasks as usize)
                         .map(|i| {
                             build_fuzzer_sub_task(
-                                &base_task_name,
+                                &multiversion_task.suite_name,
```
trahman1318, line 233: This makes it so that the execution task has the same name as the suite, which will make things more clear.

Have a failing execution task? Copy the string and search for it in the repo, you should immediately see the matrix suite file, the generated suite file, and the "initialize multiversion tasks" var name in `definitions.yml`.

Before these changes, you can search for the failing execution task string, but wont find any hits in the repo.

src/task_types/resmoke_tasks.rs
```diff
@@ -552,28 +553,22 @@ impl GenResmokeTaskServiceImpl {
     /// # Returns
     ///
     /// List of sub-suites that includes versions fall all multiversion combinations.
-    async fn create_multiversion_combinations(
+    async fn create_multiversion_tasks(
         &self,
         params: &ResmokeGenParams,
         build_variant: &str,
     ) -> Result<Vec<SubSuite>> {
+        if params.multiversion_generate_tasks.is_empty() {
```
trahman1318, line 561: Same as above for `fuzzer_tasks.rs`.

trahman1318 (commented):
Some notes for me to remember.

trahman1318:
TODO:
- Check the actual diff & make sure things look very close to the same -- especially for `evergreen_config.json`
    - Checked this & things look good so far. But will reconfirm after updates.
- ~~Fix up burn_in_tests.rs (should be the same as what was done for `resmoke_tasks.rs` & `fuzzer_tasks.rs`~~
- Fix any broken tests & write any new tests (probably a couple unit tests & maybe an integration test)
- Update docstrings/docs
- ~~Remove dead code from the now outdated `MultiversionService`~~

tests/integration_test.rs
```diff
@@ -31,7 +31,7 @@ fn test_end2end_execution() {
     assert!(tmp_dir_path.exists());
 
     let files = std::fs::read_dir(tmp_dir_path).unwrap();
-    assert_eq!(846, files.into_iter().collect::<Vec<_>>().len());
+    assert_eq!(686, files.into_iter().collect::<Vec<_>>().len());
```
trahman1318, line 34: This is because the mock integration test infra was automatically considering every multiversion suite as a ReplicaSet, which would mean every multiversion task would generate many tasks. Now that the generated tasks are explicitly written in the evergreen.yml, that "mock" codepath is not used & the number of files is significantly less.

trahman1318:
We walked through this PR last week & @zamj mentioned we need to know the behavior wrt older branches which use the task generator. The task generator goes as far back as v6.0 so I did a trial backport just to make sure this works. Looks like this should work.

PB: https://spruce.mongodb.com/version/64482c0cd6d80abf10970156/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

There is a failing `version_burn_in_gen` which I'm looking into, but overall, this looks "backportable" so we should be able to continue with this review. Let me know if anything else is needed.

TODO:
- Provide data to show that the before & after of these changes are the same (regarding tasks generated)

namazgurbanovmdb:
v0.7.5 is the current version. I might need to bump it up to v0.7.6

src/task_types/fuzzer_tasks.rs
zamj: I think this sounds like a helpful change, but can you put an example here? I just would like to see what this looks like.

Also, have you synced with someone from Server regarding this? I just want to make sure there won't be any confusion or cause any issues when this changes.

zamj (commented):
I think this overall LGTM, but I'd like @MikhailShchatko to also take a look before merging.

Also, can you share your investigation into any differences between what's generated when running this new version vs the current version? I just want to be sure we're not accidentally going to be missing a bunch of coverage or also adding a bunch without knowing.

MikhailShchatko:
> Edit: Got `version_burn_in_gen` passing on this [PB](https://spruce.mongodb.com/version/6448378ce3c331b28b766553/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

Is the last PB link correct? I see that `version_burn_in_gen` still fails there.

MikhailShchatko:
Looks good overall % a couple of things.
1. [nit] Could you please create a DAG ticket for this change and update PR title to include DAG ticket instead of SERVER?
2. Could you please run another similar PB, but also making a dummy change to any JS test to trigger burn-in? I normally do something like [this](url).

src/task_types/fuzzer_tasks.rs
trahman1318: Old generated config
```json
  {
      "name": "aggregation_multiversion_fuzzer_last_lts_1-linux",
      "commands": [
        {
          "func": "git get project no modules"
        },
        {
          "func": "add git tag"
        },
        {
          "func": "do setup"
        },
        {
          "func": "configure evergreen api credentials"
        },
        {
          "func": "do multiversion setup"
        },
        {
          "func": "setup jstestfuzz"
        },
        {
          "func": "run jstestfuzz",
          "vars": {
            "npm_command": "agg-fuzzer",
            "jstestfuzz_vars": "--numGeneratedFiles 5 "
          }
        },
        {
          "func": "run generated tests",
          "vars": {
            "timeout_secs": 1800,
            "suite": "generational_fuzzer_last_lts",
            "resmoke_jobs_max": 1,
            "resmoke_args": "--mongodSetParameters='{logComponentVerbosity: {command: 2}}'",
            "gen_task_config_location": "mongodb-mongo-master/abc123/generate_tasks/generated-config-321ab.tgz",
            "require_multiversion_setup": true,
            "task": "aggregation_multiversion_fuzzer",
            "multiversion_exclude_tags_version": "last_lts",
            "continue_on_failure": false,
            "should_shuffle": false
          }
        }
      ],
      "depends_on": [
        {
          "name": "version_gen"
        },
        {
          "name": "archive_dist_test_debug"
        }
      ]
    }
```
New generated config
```json
   {
      "name": "aggregation_multiversion_fuzzer_last_lts_1-linux",
      "commands": [
        {
          "func": "git get project no modules"
        },
        {
          "func": "add git tag"
        },
        {
          "func": "do setup"
        },
        {
          "func": "configure evergreen api credentials"
        },
        {
          "func": "do multiversion setup"
        },
        {
          "func": "setup jstestfuzz"
        },
        {
          "func": "run jstestfuzz",
          "vars": {
            "npm_command": "agg-fuzzer",
            "jstestfuzz_vars": "--numGeneratedFiles 5 "
          }
        },
        {
          "func": "run generated tests",
          "vars": {
            "resmoke_jobs_max": 1,
            "gen_task_config_location": "mongodb-mongo-master/abc123/generate_tasks/generated-config-321ab.tgz",
            "timeout_secs": 1800,
            "multiversion_exclude_tags_version": "last_lts",
            "suite": "aggregation_multiversion_fuzzer_last_lts",
            "should_shuffle": false,
            "task": "aggregation_multiversion_fuzzer",
            "resmoke_args": "--mongodSetParameters='{logComponentVerbosity: {command: 2}}'",
            "require_multiversion_setup": true,
            "continue_on_failure": false
          }
        }
      ],
      "depends_on": [
        {
          "name": "version_gen"
        },
        {
          "name": "archive_dist_test_debug"
        }
      ]
    }
```

If you look at this section, you'll see that in the new config, the task name is identical to the suite name. Therefore, when a dev sees that task failing in evergreen it is very easy to determine search and find the suite. In the old config you will see that the task name and suite name differ. For a dev to find the underlying suite that is running in the task, it requires a lot of digging.
```json
{
          "func": "run generated tests",
          "vars": {
            ...
            "suite": "...",
            ...
          }
        }
```

Regarding syncing with SERVER, there should be no noticable changes on their side of things as a result of these changes. But, a new requirement of this API is that each `multiversion_task.suite_name` must be unique (this is because we now use the `multiversion_task.suite_name` as the task name & evergreen requires unique task names). More context [here](https://github.com/10gen/mongo/pull/11789#discussion_r1172939305). I am syncing with a few people on SERVER on that.

trahman1318:
> Also, can you share your investigation into any differences between what's generated when running this new version vs the current version? I just want to be sure we're not accidentally going to be missing a bunch of coverage or also adding a bunch without knowing.

Sure, I wrote a script to check the differences between the two generated configs and the only difference was the names of a couple of tasks. Other than that, the number of tasks generated and what is actually generated is identical.

trahman1318:
> Looks good overall % a couple of things.
> 
> 1. [nit] Could you please create a DAG ticket for this change and update PR title to include DAG ticket instead of SERVER?
> 2. Could you please run another similar PB, but also making a dummy change to any JS test to trigger burn-in? I normally do something like [this](https://evergreen.mongodb.com/filediff/644a35fa0305b92eac6d32f9/?file_name=jstests%2Fcore%2Fquery%2Fall%2Fall.js&patch_number=0).

1. I promise I'll do that before merge
2. https://spruce.mongodb.com/version/644ac2c69ccd4e0d740b89c4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
^ Waiting on the results of the PB

trahman1318 (commented):
TODO:
- Docs update
- Create DAG ticket

MikhailShchatko (approved):
LGTM assuming promises are fulfilled 😄

zamj (approved):
LGTM

src/lib.rs
```diff
@@ -174,6 +174,7 @@ impl Dependencies {
             .get_generate_sub_tasks_config()?;
         let config_extraction_service = Arc::new(ConfigExtractionServiceImpl::new(
             evg_config_utils.clone(),
+            multiversion_service.clone(),
```
trahman1318, line 177: Start passing in a `multiversion_service` instance to our "config extractor".

src/services/config_extraction.rs
```diff
@@ -189,7 +192,11 @@ impl ConfigExtractionService for ConfigExtractionServiceImpl {
             require_multiversion_setup: evg_config_utils
                 .get_task_tags(task_def)
                 .contains(MULTIVERSION),
-            multiversion_generate_tasks: evg_config_utils.get_multiversion_generate_tasks(task_def),
+            multiversion_generate_tasks: self
+                .multiversion_service
+                .filter_multiversion_generate_tasks(
+                    evg_config_utils.get_multiversion_generate_tasks(task_def),
+                ),
```
trahman1318, line 199: Use the `multiversion_service` to filter `multiversion_generate_tasks` so that only the relevant tasks are returned.

src/task_types/multiversion.rs
```diff
@@ -94,4 +109,113 @@ impl MultiversionService for MultiversionServiceImpl {
...
+    /// Get the filtered multiversion generate tasks based on the config.
+    ///
+    /// # Arguments
+    ///
+    /// * `multiversion_generate_tasks` - Array of desired tasks to generate.
+    ///
+    /// # Returns
+    ///
+    /// Multiversion generate tasks that match the config's `last versions`.
+    fn filter_multiversion_generate_tasks(
```
trahman1318, line 122: `multiversion_config` is the output from running `buildscripts/resmoke.py multiversion-config`.
ie:
```
last_continuous_fcv: '7.0'
last_lts_fcv: '7.0'
last_versions:
- last_lts
requires_fcv_tag: requires_fcv_71
requires_fcv_tag_continuous: requires_fcv_71
requires_fcv_tag_lts: requires_fcv_71
```
When last-lts == last-continuous only last-lt is present in `last_versions`. So we filter the `multiversion_generate_tasks` and make sure that each multiversion task's "old version" exists in `last_versions`

trahman1318 (commented):
The old task generator would only generate `last-lts` tasks whenever `last-lts` == `last-continuous`. The new task generator did not have this behavior so I added it in this commit: https://github.com/mongodb/mongo-task-generator/pull/59/commits/40c2bea7b30e8eaa6d61bb213fa49b3d503fa8f4

PTAL @zamj @MikhailShchatko -- sorry in advance for the late commit

MikhailShchatko:
> The old task generator would only generate `last-lts` tasks whenever `last-lts` == `last-continuous`. The new task generator did not have this behavior so I added it in this commit: [40c2bea](https://github.com/mongodb/mongo-task-generator/commit/40c2bea7b30e8eaa6d61bb213fa49b3d503fa8f4)
> 
> PTAL @zamj @MikhailShchatko -- sorry in advance for the late commit
> 
> Old task-gen PB: https://spruce.mongodb.com/version/64597fe83066155df8b173e0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC New task-gen PB: https://spruce.mongodb.com/version/64597f933066155df8b17279/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

LGTM.

P.S. I guess we just have to keep some logic in task generator that will not be reflected in the evergreen.yml config. But this case doesn't seem to be super odd or confusing, so I think we should be good.

trahman1318:
Just checked the 2 PBs from above and this does not look like it has any failures or loss in coverage. Going to merge.

src/services/config_extraction.rs
```diff
@@ -217,7 +225,7 @@ impl ConfigExtractionService for ConfigExtractionServiceImpl {
         let suite = self.evg_config_utils.find_suite_name(task_def).to_string();
         let task_tags = self.evg_config_utils.get_task_tags(task_def);
         let require_multiversion_setup = task_tags.contains(MULTIVERSION);
-        let no_multiversion_iteration = task_tags.contains(NO_MULTIVERSION_ITERATION);
+        let no_multiversion_generate_tasks = task_tags.contains(NO_MULTIVERSION_GENERATE_TASKS);
```
trahman1318, line 228: @stevegrossmongodb, this is how I propagated a "tag" from the source evergreen task config to a `ResmokeGenParams` instance.

PR #64, opened 2023-05-17
DAG-2605 Accommodate required and suggested variants for burn in tags

trahman1318:
Corresponding mongo PR: https://github.com/10gen/mongo/pull/12975

### DIFF
--- a/src/evergreen/evg_config.rs
+++ b/src/evergreen/evg_config.rs
@@ -4,7 +4,7 @@ use std::{collections::HashMap, path::Path, process::Command};
 
 use shrub_rs::models::{project::EvgProject, task::EvgTask, variant::BuildVariant};
 
-const REQUIRED_PREFIX: &str = "-required";
+const REQUIRED_PREFIX: &str = "!";
 
 pub trait EvgConfigService: Sync + Send {
     /// Get a map of build variant names to build variant definitions.
@@ -57,10 +57,11 @@ impl EvgConfigService for EvgProjectConfig {
     fn sort_build_variants_by_required(&self) -> Vec<String> {
         let build_variant_map = self.get_build_variant_map();
         let mut build_variants: Vec<String> = build_variant_map
-            .keys()
-            .filter_map(|bv| {
-                if bv.ends_with(REQUIRED_PREFIX) {
-                    Some(bv.to_string())
+            .iter()
+            .filter_map(|(name, build_variant)| {
+                let display_name = build_variant.display_name.as_ref().unwrap();
+                if display_name.starts_with(REQUIRED_PREFIX) {
+                    Some(name.to_string())
                 } else {
                     None
                 }
@@ -69,10 +70,11 @@ impl EvgConfigService for EvgProjectConfig {
 
         build_variants.extend::<Vec<String>>(
             build_variant_map
-                .keys()
-                .filter_map(|bv| {
-                    if !bv.ends_with(REQUIRED_PREFIX) {
-                        Some(bv.to_string())
+                .iter()
+                .filter_map(|(name, build_variant)| {
+                    let display_name = build_variant.display_name.as_ref().unwrap();
+                    if !display_name.starts_with(REQUIRED_PREFIX) {
+                        Some(name.to_string())
                     } else {
                         None
                     }
--- a/src/evergreen/evg_config_utils.rs
+++ b/src/evergreen/evg_config_utils.rs
@@ -9,8 +9,9 @@ use shrub_rs::models::params::ParamValue;
 use shrub_rs::models::{commands::FunctionCall, task::EvgTask, variant::BuildVariant};
 
 use crate::evergreen_names::{
-    ENTERPRISE_MODULE, GENERATE_RESMOKE_TASKS, INITIALIZE_MULTIVERSION_TASKS, IS_FUZZER, LINUX,
-    MACOS, RUN_RESMOKE_TESTS, WINDOWS,
+    BURN_IN_TAG_EXCLUDE_BUILD_VARIANTS, BURN_IN_TAG_INCLUDE_BUILD_VARIANTS,
+    BURN_IN_TAG_REQUIRED_AND_SUGGESTED_BUILD_VARIANTS, ENTERPRISE_MODULE, GENERATE_RESMOKE_TASKS,
+    INITIALIZE_MULTIVERSION_TASKS, IS_FUZZER, LINUX, MACOS, RUN_RESMOKE_TESTS, WINDOWS,
 };
 use crate::utils::task_name::remove_gen_suffix;
 
@@ -165,6 +166,22 @@ pub trait EvgConfigUtils: Sync + Send {
         &self,
         name: &str,
         build_variant: &BuildVariant,
+    ) -> Vec<String>;
+
+    /// Determine corresponding burn in tag build variants for the given build variant.
+    ///
+    /// # Arguments
+    ///
+    /// * `build_variant` - Build Variant to query.
+    /// * `build_variant_map` - A map of build variant names to their definitions.
+    ///
+    /// # Returns
+    ///
+    /// List of build variant names to use for burn in tags
+    fn resolve_burn_in_tag_build_variants(
+        &self,
+        build_variant: &BuildVariant,
+        build_variant_map: &HashMap<String, &BuildVariant>,
     ) -> Vec<String>;
 
     /// Lookup the given variable in the task definition.
@@ -517,6 +534,61 @@ impl EvgConfigUtils for EvgConfigUtilsImpl {
             .collect()
     }
 
+    /// Determine burn in tag build variants for the given build variant.
+    ///
+    /// # Arguments
+    ///
+    /// * `build_variant` - Build Variant to query.
+    /// * `build_variant_map` - A map of build variant names to their definitions.
+    ///
+    /// # Returns
+    ///
+    /// List of build variant names to use for burn in tags.
+    fn resolve_burn_in_tag_build_variants(
+        &self,
+        build_variant: &BuildVariant,
+        build_variant_map: &HashMap<String, &BuildVariant>,
+    ) -> Vec<String> {
+        let mut burn_in_build_variants = self
+            .lookup_and_split_by_whitespace_build_variant_expansion(
+                BURN_IN_TAG_INCLUDE_BUILD_VARIANTS,
+                build_variant,
+            );
+        if self
+            .lookup_build_variant_expansion(
+                BURN_IN_TAG_REQUIRED_AND_SUGGESTED_BUILD_VARIANTS,
+                build_variant,
+            )
+            .unwrap_or_default()
+            == "true"
+        {
+            burn_in_build_variants.extend(
+                build_variant_map
+                    .iter()
+                    .filter_map(|(name, build_variant)| {
+                        let display_name = build_variant.display_name.as_ref().unwrap();
+                        if display_name.starts_with('!') || display_name.starts_with('*') {
+                            Some(name.to_string())
+                        } else {
+                            None
+                        }
+                    })
+                    .collect::<Vec<String>>(),
+            );
+        }
+        let exclude_burn_in_build_variants = self
+            .lookup_and_split_by_whitespace_build_variant_expansion(
+                BURN_IN_TAG_EXCLUDE_BUILD_VARIANTS,
+                build_variant,
+            );
+        burn_in_build_variants
+            .into_iter()
+            .collect::<HashSet<String>>()
+            .into_iter()
+            .filter(|name| !exclude_burn_in_build_variants.contains(name))
+            .collect::<Vec<String>>()
+    }
+
     /// Lookup the given variable in the task definition.
     ///
     /// # Arguments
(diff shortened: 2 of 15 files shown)


trahman1318:
Ready for first pass.
PB: https://spruce.mongodb.com/version/6464208932f417f987207e64/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

MikhailShchatko (approved):
LGTM

src/task_types/burn_in_tests.rs
```diff
@@ -411,7 +411,7 @@ impl BurnInService for BurnInServiceImpl {
         gen_config.build_variant_display_name = base_build_variant
             .display_name
             .as_ref()
-            .map(|s| format!("! {}", s));
+            .map(|s| format!("  JSTESTS_AFFECTED {}", s));
```
IamXander, line 414: Should this be a const string as well?

IamXander (approved):

trahman1318:
Minimal changes in the latest commit: https://github.com/mongodb/mongo-task-generator/pull/64/commits/24cd95edc78cae536118f299aa2c2e7e23e661e8

To match mongo behavior from the recent commit on the other PR: https://github.com/10gen/mongo/pull/12975/commits/ae51ee98c01099d45e4b9ebbebcea02b5774d294

The basic idea is to cache test membership so that we don't have to recalculate it on every burn-in invocation from the task-generator. Determining test membership is a time consuming step that we only need to do 1 time. This task-gen step went from ~15 min to ~1 min, which will be much better for developer experience.

zamj:
This is being incredibly nit picky, but right now the new variants wind up coming first in the list on the left side of the patch build in Spruce. Do we want that? Do we want "regular" variants to come first then these come after them? How do server engineers want to be able to reason about what variants are what?

If you've already spoke to people about this, feel free to ignore this. I can see both ways on this, I just wanted to raise the question. Changes overall LGTM though

trahman1318:
> Awesome job on the speedup! That should save a lot of time (and get some nice savings). We run this for almost every patch, so 14min saved on every patch should add up over time. 

Thanks! Gotta say though that this was when running locally. In CI, the saving are less, but it is still about a 50% save.

> right now the new variants wind up coming first in the list on the left side of the patch build in Spruce. Do we want that?

This is definitely a valid question. We discussed this & concluded that we want this surfaced at the top so devs know the exact impact of their changes and "have no excuse" for committing a failing jstest on required/suggested variants to master. I'm open to ideas here though.

Another thing to note is I've been playing around with [`! ! ! JSTESTS AFFECTED ...`](https://spruce.mongodb.com/version/6464208932f417f987207e64/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) vs. [`  JSTESTS AFFECTED ...`](https://spruce.mongodb.com/version/646537f92fbabe4137d86d2f/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&statuses=success&variant=%5Elinux-x86-dynamic-grpc-suggested%24). Both will push the variants to the top. Any preferences? @IamXander @zamj @MikhailShchatko

IamXander:
I think it should be at the same levels as the other variants and just go wherever in the alphabet it lands

So task-gen is significantly faster - maybe it is worth a quick query to compute cost savings

trahman1318:
> So task-gen is significantly faster - maybe it is worth a quick query to compute cost savings

TLDR: We actually made the mongo-task-gen step of version_burn_in_gen go from [30 seconds](https://evergreen.mongodb.com/lobster/evergreen/task/mongodb_mongo_master_generate_tasks_for_version_version_burn_in_gen_a723af8863c5fae1eee7b0a891066e923468e974_23_05_18_12_10_04/0/task#bookmarks=0%2C369&l=1&shareLine=116) to [~1 min](https://evergreen.mongodb.com/lobster/evergreen/task/mongodb_mongo_master_generate_tasks_for_version_version_burn_in_gen_patch_9f0e61a4264cfab44af0090252f6c32c6562afd4_646537f92fbabe4137d86d2f_23_05_17_20_24_42/0/task#bookmarks=0%2C758&l=1&shareLine=507), but without caching, we would have blown this out to [~10 mins](https://evergreen.mongodb.com/lobster/evergreen/task/mongodb_mongo_master_generate_tasks_for_version_version_burn_in_gen_patch_9f0e61a4264cfab44af0090252f6c32c6562afd4_6464a9f0850e610c797e98d0_23_05_17_10_18_53/0/task#bookmarks=0%2C750&l=1&shareLine=499)

I mentioned on the [other PR](https://github.com/10gen/mongo/pull/12975#issuecomment-1551882339) 

> This task-gen step went from [554 seconds](https://evergreen.mongodb.com/lobster/evergreen/task/mongodb_mongo_master_generate_tasks_for_version_version_burn_in_gen_patch_9f0e61a4264cfab44af0090252f6c32c6562afd4_6464a9f0850e610c797e98d0_23_05_17_10_18_53/0/task#bookmarks=0%2C750&l=1&shareLine=499) to [71 seconds](https://evergreen.mongodb.com/lobster/evergreen/task/mongodb_mongo_master_generate_tasks_for_version_version_burn_in_gen_patch_9f0e61a4264cfab44af0090252f6c32c6562afd4_64651ae90305b9f30a28ca56_23_05_17_18_20_34/0/task#bookmarks=0%2C756&l=1&shareLine=503)

This only affects `version_burn_in_gen` not `version_gen` so this is going to have less impact. Also, the improvement via caching is a solution to a problem surfaced in this PR. The huge increase in the number of burn_in build variants made this go from 3.5 mins ([version_burn_in_gen task history](https://spruce.mongodb.com/variant-history/mongodb-mongo-master/generate-tasks-for-version)) to [15 mins](https://spruce.mongodb.com/task/mongodb_mongo_master_generate_tasks_for_version_version_burn_in_gen_patch_9f0e61a4264cfab44af0090252f6c32c6562afd4_6464a9f0850e610c797e98d0_23_05_17_10_18_53/logs?execution=0). The caching solution brings this back down to [8 mins](https://spruce.mongodb.com/task/mongodb_mongo_master_generate_tasks_for_version_version_burn_in_gen_patch_9f0e61a4264cfab44af0090252f6c32c6562afd4_64651ae90305b9f30a28ca56_23_05_17_18_20_34/logs?execution=0) for most runs that I see. It looks like on the 8 min runs, theres a [5 min long venv setup](https://evergreen.mongodb.com/lobster/evergreen/task/mongodb_mongo_master_generate_tasks_for_version_version_burn_in_gen_patch_9f0e61a4264cfab44af0090252f6c32c6562afd4_646537f92fbabe4137d86d2f_23_05_17_20_24_42/0/task#bookmarks=0%2C758&l=1&shareLine=86) -- I'm guessing this is because in the waterfall a cached venv is being used, while on these PBs a new one is being created.

src/task_types/burn_in_tests.rs
trahman1318: Good call. Done.

src/evergreen/evg_config.rs
trahman1318, line 1: Drive by to use "!" instead of "-required"

src/evergreen/evg_config_utils.rs
```diff
@@ -517,6 +534,61 @@ impl EvgConfigUtils for EvgConfigUtilsImpl {
...
+    ///
+    /// # Returns
+    ///
+    /// List of build variant names to use for burn in tags.
+    fn resolve_burn_in_tag_build_variants(
+        &self,
+        build_variant: &BuildVariant,
+        build_variant_map: &HashMap<String, &BuildVariant>,
+    ) -> Vec<String> {
+        let mut burn_in_build_variants = self
```
trahman1318, line 552: Get "includes"

src/evergreen/evg_config_utils.rs
```diff
@@ -517,6 +534,61 @@ impl EvgConfigUtils for EvgConfigUtilsImpl {
...
+                BURN_IN_TAG_INCLUDE_BUILD_VARIANTS,
+                build_variant,
+            );
+        if self
+            .lookup_build_variant_expansion(
+                BURN_IN_TAG_REQUIRED_AND_SUGGESTED_BUILD_VARIANTS,
+                build_variant,
+            )
+            .unwrap_or_default()
+            == "true"
```
trahman1318, line 563: If `BURN_IN_TAG_REQUIRED_AND_SUGGESTED_BUILD_VARIANTS` is set, we extend "includes" with all required and suggested.

src/evergreen/evg_config_utils.rs
```diff
@@ -517,6 +534,61 @@ impl EvgConfigUtils for EvgConfigUtilsImpl {
...
+                        if display_name.starts_with('!') || display_name.starts_with('*') {
+                            Some(name.to_string())
+                        } else {
+                            None
+                        }
+                    })
+                    .collect::<Vec<String>>(),
+            );
+        }
+        let exclude_burn_in_build_variants = self
```
trahman1318, line 579: Get "excludes"

src/evergreen/evg_config_utils.rs
```diff
@@ -517,6 +534,61 @@ impl EvgConfigUtils for EvgConfigUtilsImpl {
...
+            .lookup_and_split_by_whitespace_build_variant_expansion(
+                BURN_IN_TAG_EXCLUDE_BUILD_VARIANTS,
+                build_variant,
+            );
+        burn_in_build_variants
```
trahman1318, line 584: Resolves to the unique build variants to "burn in"

src/evergreen/evg_config_utils.rs
```diff
@@ -748,15 +820,156 @@ fn get_resmoke_vars(task: &EvgTask) -> Option<&HashMap<String, ParamValue>> {
...
     use maplit::hashmap;
     use rstest::rstest;
     use shrub_rs::models::commands::{fn_call, fn_call_with_params};
     use shrub_rs::models::params::ParamValue;
+    use shrub_rs::models::project::EvgProject;
     use shrub_rs::models::task::TaskDependency;
 
     use super::*;
 
+    // test burn in variant resolver
```
trahman1318, line 835: Tests from this point forward.

src/evergreen_names.rs
```diff
@@ -93,8 +93,13 @@ pub const MULTIVERSION_EXCLUDE_TAGS: &str = "multiversion_exclude_tags_version";
 // Build Variant expansions.
 /// Name of large distro for build variant.
 pub const LARGE_DISTRO_EXPANSION: &str = "large_distro_name";
-/// List of build variant names delimited by spaces to generate burn_in_tags for.
-pub const BURN_IN_TAG_BUILD_VARIANTS: &str = "burn_in_tag_buildvariants";
+/// List of build variant names delimited by spaces to generate burn_in_tests for.
+pub const BURN_IN_TAG_INCLUDE_BUILD_VARIANTS: &str = "burn_in_tag_include_build_variants";
```
trahman1318, line 97: This field is functionally the same. It has just been renamed to accommodate the new API.

src/evergreen/evg_config_utils.rs
MikhailShchatko, line 563: Comparing string value with "true" seems a bit sketchy. I'm not super familiar with type conversion in rust, maybe [this](https://doc.rust-lang.org/std/string/struct.String.html#method.parse) can help.

src/evergreen_names.rs
```diff
@@ -93,8 +93,13 @@ pub const MULTIVERSION_EXCLUDE_TAGS: &str = "multiversion_exclude_tags_version";
...
+/// Generate burn_in_tests for all required and suggested build variants.
+pub const BURN_IN_TAG_REQUIRED_AND_SUGGESTED_BUILD_VARIANTS: &str =
+    "burn_in_tag_required_and_suggested_build_variants";
```
MikhailShchatko, line 100: It took a while for me to understand that this expansion should hold a boolean value. Can we rename it to `burn_in_tag_include_all_required_and_suggested_build_variants`? I think this way it will be more clear.

MikhailShchatko (commented):
Sorry for delayed response. Recent changes look great. Added some comments to improve.

src/evergreen/evg_config_utils.rs
```diff
@@ -517,6 +534,61 @@ impl EvgConfigUtils for EvgConfigUtilsImpl {
...
+            )
+            .unwrap_or_default()
+            == "true"
+        {
+            burn_in_build_variants.extend(
+                build_variant_map
+                    .iter()
+                    .filter_map(|(name, build_variant)| {
+                        let display_name = build_variant.display_name.as_ref().unwrap();
+                        if display_name.starts_with('!') || display_name.starts_with('*') {
```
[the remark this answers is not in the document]
trahman1318: We could add it here in `evg_config_utils.rs` but the way things are currently designed, it could only be used here. We wouldn't easily be able to reuse it in `evg_config.rs` at the [other place](https://github.com/trahman1318/mongo-task-generator/blob/abc85f6f108d451330ff05d8fb7319f2a84a8aad/src/evergreen/evg_config.rs#L63). That brings up some questions about should a `EvgProjectConfig` struct from `evg_project.rs` "have a" `EvgConfigUtils` from `evg_config_utils.rs`, etc.

It seems to me like these are intentionally decoupled but I could be overthinking it. Maybe the real solution would be to put some of this logic into [shrub-rs](https://github.com/dbradf/shrub-rs/blob/b2944d31269d2fc69bca0b4acc2398696f998fb6/src/models/variant.rs#LL13C6-L13C6)? Anyways, I'd like to leave this for now & if we need to I think we can pull this out in the future.

src/evergreen_names.rs
trahman1318: Yeah, makes sense. Updated.

src/evergreen/evg_config_utils.rs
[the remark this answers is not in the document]
MikhailShchatko: Moving this logic to `shrub-rs` seems reasonable

zamj (approved):
Overall LGTM

PR #66, opened 2023-05-25
DAG-2605: Negiligible change to trigger/fix release pipeline

trahman1318:


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 0.7.8 - 2022-05-25
```
namazgurbanovmdb, line 3: Was this the only place that was missed? I believe the version should be updated in more files.

trahman1318: The "real" v0.7.8 code & changes are in [this commit](https://github.com/mongodb/mongo-task-generator/commit/6028dafbf230ec1bf519abded111157c504fcf0f). This is just a follow-up commit.

This PR doesn't do anything and is a potential workaround. I'm stuck in a weird spot where I can't generate a new  mongo-task-generator release for some reason.

I have [this commit](https://github.com/mongodb/mongo-task-generator/commit/6028dafbf230ec1bf519abded111157c504fcf0f) which is the "real commit" & was approved by DAG.

I then realized I had the change log wrong, so I made [this followup commit](https://github.com/mongodb/mongo-task-generator/commit/5d9a26f7ef77adb6e6b9aad38c70b9860de53db8). Since it was just a one line doc change, I just force merged with the "squash & merge button". After that commit, I have been unable to create a new release.
image

I get the following error:
https://github.com/mongodb/mongo-task-generator/actions/runs/5084796697/jobs/9137591223
image

I tried using different tag names that haven't been used yet, but still get the same error.

I figured it might have something to do with my force merging so I made this PR in case this is the solution. If there's another way to do this, I can try that too. I'm open to ideas.

IamXander (approved):

CHANGELOG.md
trahman1318: Also put some info in the slack channel but basically says the same thing: https://mongodb.slack.com/archives/C036F3YQ796/p1685051723965299

trahman1318:
Going for the merge. I think this should allow me to create a new release properly.

PR #71, opened 2024-02-16
DEVPROD-4926 add use_xlarge_distro option to mongo-task-generator

Trevor159:
for https://jira.mongodb.org/browse/BF-29398 we need to increase the distro size of one generated task but currently generated tasks only allow the variant distro or the configured large distro for a variant. This ticket will add a third option so we will not have to increase the distro size for every task unnecessarily. 

Here are the server changes that are necessary for this https://github.com/10gen/mongo/pull/19089

use_xlarge_distro acts as use_large_distro for variants that do not have xlarge_distro_name configured.

### DIFF
--- a/src/services/config_extraction.rs
+++ b/src/services/config_extraction.rs
@@ -427,6 +427,7 @@ mod tests {
                         ..Default::default()
                     },
                     use_large_distro: *value,
+                    use_xlarge_distro: false,
                 })
                 .collect(),
         };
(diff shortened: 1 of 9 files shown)


IamXander (approved):

src/services/config_extraction.rs
```diff
@@ -286,9 +292,16 @@ impl ConfigExtractionService for ConfigExtractionServiceImpl {
         let large_distro_name = self
             .evg_config_utils
             .lookup_build_variant_expansion(LARGE_DISTRO_EXPANSION, build_variant);
+        let xlarge_distro_name = self
+            .evg_config_utils
+            .lookup_build_variant_expansion(XLARGE_DISTRO_EXPANSION, build_variant);
         let build_variant_name = build_variant.name.as_str();
 
-        if generated_task.use_large_distro() {
+        if generated_task.use_xlarge_distro() && xlarge_distro_name.is_some() {
```
dmoody256, line 300: what if you specify both large and xlarge distros? Does that work and it looks like xlarge would take precedence?

Trevor159: Yes, it works fine and xlarge takes precedence which I think is desired behavior.

PR #72, opened 2024-02-29
DEVPROD-5087 allow variants to override last_versions

Trevor159:
Rust noob

Our git future multiversion variant would not run last continuous tests when last_lts = last_continuous. This allows variants to override the `last_versions` setting in the multiversion_config so they will always have the versions they want.

The UNIQUE_GEN_SUFFIX_EXPANSION is needed so we do not have duplicate tasks with the same name because there can now be conflicts.

Here is a PB on the mongodb-mongo-master project with this implemented: https://spruce.mongodb.com/version/65de6b463e8e8682d69db4f4

The mongo changes for this are here https://github.com/10gen/mongo/pull/19452

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -787,14 +787,14 @@ fn lookup_task_name(
             task_name,
             platform,
             ENTERPRISE_MODULE,
-            unique_gen_suffix.unwrap_or("".to_string())
+            unique_gen_suffix.as_deref().unwrap_or("")
         )
     } else {
         format!(
             "{}-{}{}",
             task_name,
             platform,
-            unique_gen_suffix.unwrap_or("".to_string())
+            unique_gen_suffix.as_deref().unwrap_or("")
         )
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -495,7 +495,7 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
                 {
                     continue;
                 }
-                let last_versions_expansion = self
+                let gen_task_suffix = self
                     .evg_config_utils
                     .lookup_build_variant_expansion(UNIQUE_GEN_SUFFIX_EXPANSION, build_variant);
                 // Skip tasks that have already been seen.
@@ -503,7 +503,7 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
                     is_enterprise,
                     &task.name,
                     &platform,
-                    last_versions_expansion.as_deref(),
+                    gen_task_suffix.as_deref(),
                 );
                 if seen_tasks.contains(&task_name) {
                     continue;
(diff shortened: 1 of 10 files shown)


src/lib.rs
```diff
@@ -759,11 +775,27 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
 /// # Returns
 ///
 /// Name to use for task.
-fn lookup_task_name(is_enterprise: bool, task_name: &str, platform: &str) -> String {
+fn lookup_task_name(
+    is_enterprise: bool,
+    task_name: &str,
+    platform: &str,
+    unique_gen_suffix: Option<String>,
```
zackwintermdb, line 782: Is it possible change this to Option<&str>? I think you can deref in the calling code that invokes this function. String is generally when you want to own it rather than just reference it (but I haven't used rust in a while so maybe it's necessary here)

zackwintermdb (approved):
LGTM

zackwintermdb (approved):

src/lib.rs
```diff
@@ -495,9 +495,16 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
                 {
                     continue;
                 }
-
+                let last_versions_expansion = self
+                    .evg_config_utils
+                    .lookup_build_variant_expansion(UNIQUE_GEN_SUFFIX_EXPANSION, build_variant);
```
MikhailShchatko, line 500: Why is it named `last_versions_expansion` if it queries `UNIQUE_GEN_SUFFIX_EXPANSION`?

src/lib.rs
```diff
@@ -683,7 +691,15 @@ impl GenerateTasksService for GenerateTasksServiceImpl {
                 } else if task.name == BURN_IN_TASKS {
                     format!("{}-{}", BURN_IN_TASKS_PREFIX, bv_name)
                 } else {
-                    lookup_task_name(is_enterprise, &task.name, &platform)
+                    let last_versions_expansion = self
+                        .evg_config_utils
+                        .lookup_build_variant_expansion(UNIQUE_GEN_SUFFIX_EXPANSION, build_variant);
```
MikhailShchatko, line 696: > Why is it named `last_versions_expansion` if it queries `UNIQUE_GEN_SUFFIX_EXPANSION`?

same here

src/lib.rs
```diff
@@ -799,7 +831,14 @@ fn create_task_worker(
 
         let is_enterprise = evg_config_utils.is_enterprise_build_variant(&build_variant);
         let platform = evg_config_utils.infer_build_variant_platform(&build_variant);
-        let task_name = lookup_task_name(is_enterprise, &task_def.name, &platform);
+        let last_versions_expansion = evg_config_utils
+            .lookup_build_variant_expansion(UNIQUE_GEN_SUFFIX_EXPANSION, &build_variant);
```
MikhailShchatko, line 835: > Why is it named `last_versions_expansion` if it queries `UNIQUE_GEN_SUFFIX_EXPANSION`?

and here

MikhailShchatko (approved):
LGTM % some comments

src/lib.rs
Trevor159: good catch

PR #73, opened 2024-04-08
SERVER-88917 Update method for determining enterprise build variants …

juangugit:
…without using modules in mongo-task-generator

We previously determined if a build variant was an enterprise module by checking for the presence of 'enterprise' in the build variant's modules parameter. Since the removal of the modules:enterprise, we need to have a new method for this determination. I am using regex to match `--enableEnterpriseTests = off` flag. 

I have modified [`prelude_mongo_task_generator.sh`](https://github.com/10gen/mongo/blob/master/evergreen/prelude_mongo_task_generator.sh) file to run tests. 

I have confirmed that we have `-enterprise` for enterprise build variants. Now
image
vs Before
image

Here is the [PB](https://spruce.mongodb.com/version/661467314dc692000702edd9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.7.13 - 2024-04-08
+* SERVER-88917 Update method for determining enterprise build variants without using modules in mongo-task-generator
+
 ## 0.7.12 - 2024-02-29
 * DEVPROD-5087 allow variants to override last_versions
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1189,7 +1189,7 @@ dependencies = [
 
 [[package]]
 name = "mongo-task-generator"
-version = "0.7.12"
+version = "0.7.13"
 dependencies = [
  "anyhow",
  "assert_cmd",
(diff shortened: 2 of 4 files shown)


thomaslangston-mongodb:
It is probably worth spot checking this change in at least three buildvariants. @stevegrossmongodb WDYT?

juangugit:
> It is probably worth spot checking this change in at least three buildvariants. @stevegrossmongodb WDYT?

I ran all required builds from the PB link above. The screenshot shows the changes to expect.

thomaslangston-mongodb:
Ah, I see the link now.

Two questions:

1) Is there anything we need to do for the two failing tasks that were passing in the base which are not known issues: https://spruce.mongodb.com/version/661467314dc692000702edd9/tasks?baseStatuses=success&page=0&sorts=STATUS%3ADESC&statuses=failed

2) Is there a community buildvariant in the Patch that shows their tasks don't have the -enterprise naming convention?

thomaslangston-mongodb (approved):
Approved

juangugit:
evergreen merge

PR #83, opened 2024-10-11
DEVPROD-11978 pass --skipTestsCoveredByMoreComplexSuites flag to resmoke when generating tasks in a patch build.

kauboy26:
…

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,6 +136,8 @@ pub struct ExecutionConfiguration<'a> {
     pub config_location: &'a str,
     /// Should burn_in tasks be generated.
     pub gen_burn_in: bool,
+    /// True if this patch is a patch build.
+    pub is_patch: bool,
     /// Command to execute burn_in_tests.
     pub burn_in_tests_command: &'a str,
     /// S3 endpoint to get test stats from.
@@ -163,7 +165,7 @@ impl Dependencies {
     /// A set of dependencies to run against.
     pub fn new(execution_config: ExecutionConfiguration) -> Result<Self> {
         let fs_service = Arc::new(FsServiceImpl::new());
-        let discovery_service = Arc::new(ResmokeProxy::new(execution_config.resmoke_command));
+        let discovery_service = Arc::new(ResmokeProxy::new(execution_config.resmoke_command, execution_config.is_patch));
         let multiversion_service = Arc::new(MultiversionServiceImpl::new(
             discovery_service.get_multiversion_config()?,
         )?);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,8 +136,8 @@ pub struct ExecutionConfiguration<'a> {
     pub config_location: &'a str,
     /// Should burn_in tasks be generated.
     pub gen_burn_in: bool,
-    /// True if this patch is a patch build.
-    pub is_patch: bool,
+    /// True if the generator should skip tests covered by more complex suites.
+    pub skip_covered_tests: bool,
     /// Command to execute burn_in_tests.
     pub burn_in_tests_command: &'a str,
     /// S3 endpoint to get test stats from.
@@ -167,7 +167,7 @@ impl Dependencies {
         let fs_service = Arc::new(FsServiceImpl::new());
         let discovery_service = Arc::new(ResmokeProxy::new(
             execution_config.resmoke_command,
-            execution_config.is_patch,
+            execution_config.skip_covered_tests,
         ));
         let multiversion_service = Arc::new(MultiversionServiceImpl::new(
             discovery_service.get_multiversion_config()?,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -165,7 +165,10 @@ impl Dependencies {
     /// A set of dependencies to run against.
     pub fn new(execution_config: ExecutionConfiguration) -> Result<Self> {
         let fs_service = Arc::new(FsServiceImpl::new());
-        let discovery_service = Arc::new(ResmokeProxy::new(execution_config.resmoke_command, execution_config.is_patch));
+        let discovery_service = Arc::new(ResmokeProxy::new(
+            execution_config.resmoke_command,
+            execution_config.is_patch
+        ));
         let multiversion_service = Arc::new(MultiversionServiceImpl::new(
             discovery_service.get_multiversion_config()?,
         )?);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -167,7 +167,7 @@ impl Dependencies {
         let fs_service = Arc::new(FsServiceImpl::new());
         let discovery_service = Arc::new(ResmokeProxy::new(
             execution_config.resmoke_command,
-            execution_config.is_patch
+            execution_config.is_patch,
         ));
         let multiversion_service = Arc::new(MultiversionServiceImpl::new(
             discovery_service.get_multiversion_config()?,
--- a/src/main.rs
+++ b/src/main.rs
@@ -36,6 +36,10 @@ struct EvgExpansions {
     pub is_patch: bool,
 }
 
+// When running in a patch build, the is_patch YAML field is set to the
+// string "true" rather than a boolean `true`. Therefore we need a custom
+// deserializer to convert from the string "true" to the boolean `true`, and
+// in all other cases return `false`.
 fn deserialize_is_patch<'de, D>(deserializer: D) -> Result<bool, D::Error>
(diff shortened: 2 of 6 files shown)


src/main.rs
```diff
@@ -31,6 +31,9 @@ struct EvgExpansions {
     pub task_name: String,
     /// ID of Evergreen version running.
     pub version_id: String,
+    /// True if the patch is a patch build.
+    #[serde(default)]
```
kauboy26, line 35: I added the `serde(default)` bit in case the `is_patch` variable isn't provided. I ran the task-generator locally with and without the `is_patch` variable in the expansions file and it seemed to accordingly include / omit the `--skipTestsCoveredByMoreComplexSuite` flag when calling Resmoke.

kauboy26 (commented):
Here is an [Evergreen patch](https://spruce.mongodb.com/version/67098e46c1499100076cadf0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) of my changes - I expect the `is_patch` field to exist in the expansions file since I'm running a patch build.

I also am running a [second Evg patch](https://spruce.mongodb.com/version/670990beaec3160007a92e56/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) where I changed the `is_patch` variable to be called `is_patch_without`, which obviously won't be written in the Evg expansions file to see how it behaves. [Here](https://github.com/mongodb/mongo-task-generator/pull/84) is how I made that change.

src/main.rs
```diff
@@ -31,6 +31,20 @@ struct EvgExpansions {
     pub task_name: String,
     /// ID of Evergreen version running.
     pub version_id: String,
+    /// True if the patch is a patch build.
+    #[serde(default, deserialize_with = "deserialize_is_patch")]
+    pub is_patch: bool,
+}
+
+fn deserialize_is_patch<'de, D>(deserializer: D) -> Result<bool, D::Error>
```
kauboy26, line 39: I had to add this because the `is_patch` variable is unfortunately not a boolean, but a string. This is to deserialize that string into a boolean.

src/lib.rs
```diff
@@ -163,7 +165,7 @@ impl Dependencies {
     /// A set of dependencies to run against.
     pub fn new(execution_config: ExecutionConfiguration) -> Result<Self> {
         let fs_service = Arc::new(FsServiceImpl::new());
-        let discovery_service = Arc::new(ResmokeProxy::new(execution_config.resmoke_command));
+        let discovery_service = Arc::new(ResmokeProxy::new(execution_config.resmoke_command, execution_config.is_patch));
```
juangugit, line 168: could you double check the linting and testing [here](https://github.com/mongodb/mongo-task-generator)?

src/main.rs
MikhailShchatko: From [evg docs](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files#default-expansions)
> ${is_patch} is "true" if the running task is in a patch build and undefined if it is not.

Is it going to work as expected outside of patch buids?

src/lib.rs
kauboy26: Looks like I missed making this change - I've done it now!

Btw the [README](https://github.com/mongodb/mongo-task-generator?tab=readme-ov-file#lintingformatting) doesn't really explain how to run lint / format locally. The suggested commands don't work:
```
ubuntu@ip-10-122-6-249:~/mongo-task-generator$ cargo fmt
error: no such command: `fmt`

	Did you mean `fix`?

	View all installed commands with `cargo --list`
	Find a package to install `fmt` with `cargo search cargo-fmt`
```

src/main.rs
kauboy26: I've added a [`#[serde(default)]` above](https://github.com/mongodb/mongo-task-generator/pull/83/files/3028104ee63e3d6d37a55be8fb4b435cc1bac789#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR35) - this is so that if the `is_patch` variable doesn't exist in the YAML altogether, then it defaults to `false` when it is deserialized.

The way I tested the "missing `is_patch` case" is by changing [this code](https://github.com/mongodb/mongo-task-generator/pull/83/files/3028104ee63e3d6d37a55be8fb4b435cc1bac789#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR36) to attempt to deserialize an `is_patch_without` variable, which will not exist in the Evg expansions file. You can see that the mongo-task-generator was still able to run fine in this [patch](https://spruce.mongodb.com/version/670990beaec3160007a92e56/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) but it omitted the `--useMinimalTests` option.

If the variable exists, and is set to anything other than the string `"true"`, then again it becomes false. This is done [here](https://github.com/mongodb/mongo-task-generator/pull/83/files/3028104ee63e3d6d37a55be8fb4b435cc1bac789#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR46). Only when it is set to `"true"` is it actually deserialized into the boolean with `true`. I tested this and the above as well locally, but I couldn't figure out a way to test this case on Evg.

MikhailShchatko: Thanks for more detailed explanation. Even though I did some things in Rust for this project, it's still hard to comprehend for me sometimes.

src/lib.rs
MikhailShchatko: Hm, that's weird. My `cargo` has this command, but I don't remember setting anything up for that to work.
Anyway PR checks are green so we should be good.

MikhailShchatko (approved):
LGTM

juangugit (approved):
LGTM

kauboy26:
@juangugit and @MikhailShchatko, I added renamed the `is_patch` variable in deeper layers of the code to `skip_covered_tests` to reflect the point of the variable. Additionally, I added a new variable `run_covered_tests` that people can pass in as a patch parameter in case they want to opt-out of the feature.

CC: @Trevor159

Trevor159 (approved):

visemet:
> [concurrency_sharded_multi_stmt_txn in patch with `run_covered_tests` ](https://spruce.mongodb.com/task/mongodb_mongo_master_enterprise_amazon_linux2_arm64_all_feature_flags_display_concurrency_sharded_multi_stmt_txn_patch_b61e0a70d43ec4fa05f33fe8f2107f8d9794c0bd_6712b22ccfe9430007973707_24_10_18_19_12_24/tests?execution=0&sorts=STATUS%3AASC) runs 1603 tests whereas in a [patch without `run_covered_tests`](https://spruce.mongodb.com/task/mongodb_mongo_master_enterprise_amazon_linux2_arm64_all_feature_flags_display_concurrency_sharded_multi_stmt_txn_patch_b61e0a70d43ec4fa05f33fe8f2107f8d9794c0bd_6712b241efdbe90007b2688d_24_10_18_19_16_55/tests?execution=0&sorts=STATUS%3AASC) it runs only 127 tests.

How was the patch build with `run_covered_tests` run? Did it use `evergreen patch --param run_covered_tests=true`? Let's add something to the Evergreen project configuration so the name `run_covered_tests` is discoverable.

```
$ evergreen list --parameters -p mongodb-mongo-master
Name                             Default            Description
----                             -------            -----------
evergreen_config_file_path       etc/evergreen.yml  path to this file
patch_compile_flags                                 Additional SCons flags to be applied during scons compile invocations in this patch
future_git_tag                                      Future git tag to be added. If empty, we will use the most recent git tag instead.
last_lts_evg_version_id                             The Evergreen Version ID of the last-lts MongoDB binaries. Only binaries from release variants are used
last_continuous_evg_version_id                      The Evergreen Version ID of the last-continuous MongoDB binaries. Only binaries from release variants are used
antithesis_image_tag                                The docker tag to use when pushing images to Antithesis
build_patch_id                                      Patch id of evergreen patch to pull binaries from for testing.
project_required_suggested_cron  0 */4 * * *        Cron schedule for required and suggested variants
project_nightly_cron             0 4 * * *          Cron schedule for nightly variants
project_weekly_cron              0 4 * * 0          Cron schedule for nightly variants
bazel_scons_diff_targets                            Targets to run the bazel-scons diff on in this patch
convert_bazel_headers_target                        Target to generate bazel headers for
```

PR #87, opened 2025-01-16
DEVPROD-14170 Add additional logging for long running task generation

sleaux:


sleaux:
This logging is in support of tracking down a rare hang seen in https://jira.mongodb.org/browse/BF-35879.  

The remaining task count may support one of two suspicions:
1. If only a single task is left remaining at the hang, it is likely hung on an query to an external resource. I think this is unlikely since the web requests already timeout+retry, and in the system logs from the timed-out task, there are no resmoke.py process lingering around. 
2. If there are 2+ tasks left remaining, it's possible they are deadlocked on each other, and we can chase down that possibility further.

Patch build in mongodb-mongo-master using a mongo-task-generator built from this branch: https://spruce.mongodb.com/version/678954a36e87f90007f3969a

MikhailShchatko (approved):
Nice use of tokio lib, good job!

PR #88, opened 2025-01-22
DEVPROD-14354 avoid a division-by-zero when processing empty suites

sleaux:
The tests returned by resmoke's test discovery for a task may be empty. The codepath for splitting the task when stats from evergreen are successfully collected handles this, but the fallback method here naively did not. An empty list of sub-suites should be returned to match behavior.

### DIFF
--- a/src/task_types/resmoke_tasks.rs
+++ b/src/task_types/resmoke_tasks.rs
@@ -509,15 +509,14 @@ impl GenResmokeTaskServiceImpl {
         multiversion_name: Option<&str>,
         multiversion_tags: Option<String>,
     ) -> Result<Vec<SubSuite>> {
+        let mut sub_suites = vec![];
+
         let origin_suite = multiversion_name.unwrap_or(&params.suite_name);
         let test_list = self.get_test_list(params, multiversion_name)?;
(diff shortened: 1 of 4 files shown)


MikhailShchatko (approved):

src/task_types/resmoke_tasks.rs
```diff
@@ -512,9 +512,14 @@ impl GenResmokeTaskServiceImpl {
         let origin_suite = multiversion_name.unwrap_or(&params.suite_name);
         let test_list = self.get_test_list(params, multiversion_name)?;
         let n_suites = min(test_list.len(), self.config.n_suites);
-        let tasks_per_suite = test_list.len() / n_suites;
 
         let mut sub_suites = vec![];
+        if n_suites == 0 {
+            return Ok(sub_suites);
+        }
```
smcclure15, line 519: nit: can this early-return occur before getting `origin_suite`? Always like to push the bailouts up as high as possible

smcclure15 (approved):

src/task_types/resmoke_tasks.rs
sleaux: Not quite before `origin_suite`, but I have now moved it up as far as it can be.

PR #90, opened 2025-03-19
SERVER-102614 Add a timeout to Evergreen test stats requests

sleaux:


sleaux:
References to help review, since I wasn't familiar with these APIs when I started looking.

- `ClientBuilder` from reqwest_middleware: https://docs.rs/reqwest-middleware/latest/reqwest_middleware/
- `pub fn timeout(self, timeout: Duration) -> ClientBuilder` from reqwest: https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.timeout

lavigneer (approved):
LGTM!

PR #91, opened 2025-03-31
DEVPROD-16100 Build, test, and release on Ubuntu 24.04

sleaux:

PR #92, opened 2025-04-10
DEVPROD-16742 Include fully disabled feature flags when using resmoke's test-discovery

sleaux:
This will ensure that build variants that override fully disabled feature flags will include the tests that are expected to run.

In total, the generated suite configs in mongodb-mongo-master grow by 1.3% since they now include a few more tests, but when run, the feature flag filtering is applied as expected. Only variants that use --additionalFeatureFlags to override fully_disabled_feature_flags.yml may have a increase in what tests actually run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Changelog
+## 0.7.24 - 2025-04-10
+* Include fully disabled feature flags when using resmoke's test-discovery
+
 ## 0.7.23 - 2025-03-17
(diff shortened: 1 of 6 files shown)


sleaux:
FYI, this will precede https://github.com/mongodb/mongo-task-generator/pull/91.

Trevor159:
is it possible this will add this resmoke arg to older branches of mongo if this gets backported without the resmoke backport?

Trevor159 (approved):

PR #94, opened 2025-04-18
DEVPROD-16911 make number of sub-tasks configurable

sleaux:
This extends the capability for configuring the number of sub-tasks to all generated task types (previously this was only available for fuzzer tasks). The number of generated subtasks is now configurable using `num_tasks`, with the default still being to use 5.

Usage:
```
- <<: *gen_task_template
  name: auth_gen
  tags: ["auth"]
  commands:
  - func: "generate resmoke tasks"
    vars:
      num_tasks: 10
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1635,7 +1635,7 @@ dependencies = [
 
 [[package]]
 name = "mongo-task-generator"
-version = "1.0.0"
+version = "1.1.0"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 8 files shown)


sleaux:
In addition to the integration testing here, I also tested this in mongodb-mongo-master, splitting some generated tasks into different numbers of subtasks.

Trevor159 (approved):

PR #99, opened 2025-05-07
DEVPROD-17564 Generate multiversion binary selection tasks

sleaux:
When generating tasks, if the task is tagged with `"multiversion"`, a dependency on the new multiversion binary selection task is added.
When appending generated tasks to build variants, if any task on the build variant is multiversion, also append an instance of the multiversion binary selection task to run on that variant.


Tianyu659 (approved):
lgtm to my best knowledge 👍 you can also tag trevor/steve for this for extra caution

smcclure15 (approved):

PR #108, opened 2025-07-16
DEVPROD-19599 Generate the correct number of subtasks for Bazel-based multiversion fuzzer tasks

sleaux:


src/task_types/fuzzer_tasks.rs
```diff
@@ -319,16 +322,7 @@ fn build_fuzzer_sub_task(
     let mut run_test_vars = params.build_run_tests_vars(generated_suite_name, old_version);
 
     if params.is_bazel() {
-        let target: String = if params.is_multiversion() {
-            params.multiversion_generate_tasks.as_ref().unwrap()[sub_task_index]
-                .bazel_target
-                .clone()
-                .unwrap()
```
sleaux, line 326: The origin of the bug is here, where `sub_task_index` is _not_ the index of which multiversion task to generate.

CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # Changelog
+## 3.4.2 - 0225-07-16
```
smcclure15, line 2:
```suggestion
## 3.4.2 - 2025-07-16
```

src/task_types/fuzzer_tasks.rs
```diff
@@ -279,6 +289,7 @@ impl GenFuzzerService for GenFuzzerServiceImpl {
 /// * `params` - Parameters for how task should be generated.
 /// * `generated_suite_name` - Name of suite to execute against.
 /// * `old_version` - Previous version of mongo to test against.
+/// * `bazel_target` - Bazel target name, if running via bazel
```
smcclure15, line 292: To be more precise, is this the "short" name (eg. my_special_tests) or full name (eg. //foo/bar:my_special_tests)?

Oh - I see it's the latter in your unittest. Good to be explicit with the contract then.

smcclure15 (approved):

CHANGELOG.md
sleaux: 🤦

PR #117, opened 2026-07-15
DEVPROD-36071: Stop emitting gen_task_config_location in generated tasks

lavigneer:
## What
Stop emitting the `gen_task_config_location` var into every generated task's `run generated tests`/`run tests` function vars, and remove the now-dead `config_location` plumbing (`EvgExpansions::config_location`, `ExecutionConfiguration.config_location`, `ConfigExtractionServiceImpl` field, `ResmokeGenParams`/`FuzzerGenTaskParams` fields).

## Why
The var is an identical ~160B string repeated in all ~5,400 generated task definitions — ~1MB (measured: 0.82MB BSON, 5.6%) of the mongodb-mongo-master generated config that gets appended to the parser project document. That project crossed Evergreen's 16MB threshold, forcing S3 fallback for config fetches (DEVPROD-36071).

## Safety
No supported branch consumes the expansion: the `retrieve generated test configuration` function derives the S3 path from `${project}/${revision}/${version_id}` on master and all active release branches. Only EOL v5.0 referenced `${gen_task_config_location}`, and release branches pin older generator releases regardless.

## Testing
`cargo test` passes (integration test line counts updated); verified end-to-end on a mongodb-mongo-master patch build — task succeeded and the generated config contains zero occurrences of the var.

### DIFF
--- a/src/task_types/resmoke_tasks.rs
+++ b/src/task_types/resmoke_tasks.rs
@@ -28,9 +28,9 @@ use crate::{
     },
     evergreen_names::{
         ADD_GIT_TAG, CONFIGURE_EVG_API_CREDS, DO_MULTIVERSION_SETUP, DO_SETUP,
-        GEN_TASK_CONFIG_LOCATION, GET_PROJECT_WITH_NO_MODULES, MULTIVERSION_EXCLUDE_TAG,
-        MULTIVERSION_EXCLUDE_TAGS_FILE, REQUIRE_MULTIVERSION_SETUP, RESMOKE_ARGS, RESMOKE_JOBS_MAX,
-        RUN_GENERATED_TESTS, RUN_GENERATED_TESTS_VIA_BAZEL, SUITE_NAME,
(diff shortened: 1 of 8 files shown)


sleaux (approved):
Looks great, thanks for doing this!

src/task_types/resmoke_tasks.rs
```diff
@@ -958,7 +955,8 @@ mod tests {
 
         let test_vars = params.build_run_test_vars("my_suite_0", &sub_suite, "", None);
 
-        assert_eq!(test_vars.len(), 4);
+        assert_eq!(test_vars.len(), 3);
+        assert!(!test_vars.contains_key("gen_task_config_location"));
```
sleaux: Can you actually remove the negative `gen_task_config_location` assertions? They were great for proving completion of the removal but serve no purpose long term.

lavigneer: Done!

